PR #10, opened 2017-11-09
Add links to Android issues

lptr:


src/main/groovy/org/gradle/android/AndroidCacheFixPlugin.groovy
```diff
@@ -31,11 +31,9 @@ class AndroidCacheFixPlugin implements Plugin<Project> {
         new AndroidJavaCompile_ProcessorListFile_Workaround(),
         new ExtractAnnotations_Source_Workaround(),
         new IncrementalTask_CombinedInput_Workaround(),
-        new StreamBasedTask_CombinedInput_Workaround(),
         new ProcessAndroidResources_MergeBlameLogFolder_Workaround(),
         new CheckManifest_Manifest_Workaround(),
-    ]
-
+    ] as List<Workaround>
```
lptr: IntelliJ complains if this is not here.

wolfs: Interesting, I suggested removing it since "my" IntelliJ (2017.2.5) did not complain.

src/main/groovy/org/gradle/android/AndroidCacheFixPlugin.groovy
```diff
@@ -192,50 +193,40 @@ class AndroidCacheFixPlugin implements Plugin<Project> {
     }
 
     /**
-     * Fix {@link IncrementalTask#getCombinedInput()} relocatability.
+     * Fix {@link IncrementalTask#getCombinedInput()} and {@link StreamBasedTask#getCombinedInput()} relocatability.
      */
-    @FixedInAndroid(version = "3.0.1", link = "https://issuetracker.google.com/issues/68771542#comment3")
+    @AndroidIssue(introducedIn = "3.0.0", fixedIn = "3.0.1", link = "https://issuetracker.google.com/issues/68771542")
     static class IncrementalTask_CombinedInput_Workaround implements Workaround {
```
wolfs: I guess we should change the name to `CombinedInput_Workaround`, WDYT?

wolfs: Alternatively, we could add the same `AndroidIssue` annotation to both workaround classes.

lptr: Renamed

PR #12, opened 2017-11-13
Support Android plugin 3.1.0 alpha01

lptr:


src/test/groovy/org/gradle/android/RelocationTest.groovy
```diff
@@ -1,135 +1,375 @@
...
         def relocatedDir = temporaryFolder.newFolder()
-        new SimpleAndroidApp(relocatedDir, cacheDir, androidVersion.toString()).writeProject()
+        new SimpleAndroidApp(relocatedDir, cacheDir, androidVersion).writeProject()
 
-        def gradleVerString = gradleVersion.version
+        def expectedResults = EXPECTED_RESULTS[androidVersion]
 
-        when:
-        def result = withGradleVersion(gradleVerString)
+        println expectedResults.describe(gradleVersion)
```
wolfs: Leftover println?

src/test/groovy/org/gradle/android/RelocationTest.groovy
```diff
@@ -1,135 +1,375 @@
...
+            ':app:processDebugResources': FROM_CACHE,
+            ':app:processReleaseJavaRes': NO_SOURCE,
+            ':app:processReleaseManifest': FROM_CACHE,
+            ':app:processReleaseResources': FROM_CACHE,
+            ':app:splitsDiscoveryTaskDebug': FROM_CACHE,
+            ':app:splitsDiscoveryTaskRelease': FROM_CACHE,
+            ':app:transformClassesWithDexBuilderForDebug': SUCCESS,
+            ':app:transformClassesWithPreDexForRelease': SUCCESS,
+            // TODO For some reason this doesn't load from cache on Travis
+            ':app:transformDexArchiveWithDexMergerForDebug': { GradleVersion gradleVersion ->
```
wolfs: I think I would prefer having a method `expectedResults(androidVersion, gradleVersion)` which would then return the right `Map`. Then I could see in one place what the differences between the various Gradle versions for a fixed Android version are. Then we could get rid of the if statements checking if the value type is a `DynamicTaskOutcome`, too. WDYT?

src/test/groovy/org/gradle/android/RelocationTest.groovy
lptr: This is intentionally here, comes in handy every time something breaks on Travis.

src/test/groovy/org/gradle/android/RelocationTest.groovy
lptr: I'm not sure I understand. This way you can very easily see what happens with a fixed Android version. How would your solution look like?

wolfs: My solution would have a static map for each Android version and then the modifications to the map for a specific Gradle version in one place.
But it is fine as it is, too.

lptr: Ah, mutable, brrr! :)

PR #14, opened 2017-11-16
Support Android plugin 3.1.0-alpha02 and 03

lptr:
There are no fixes yet, only changes in task names.


lptr:
I haven't yet removed the `fixedIn = "3.0.1"` from any workarounds, but looks like we should do that at some point. Let's handle that as a separate PR, though.

PR #76, opened 2020-04-14
Add fixes for Android 3.5.x to 4.0.x cache issues

ghale:
This updates the plugin to fix issues with Android 3.5.x, 3.6.x and 4.0.x.  Specifically, it fixes the following:

- MergeJavaResourcesTask is not relocatable: https://issuetracker.google.com/issues/140602655
- MergeNativeLibsTask is not relocatable (only fixes 3.6.x and 4.0.x): https://issuetracker.google.com/issues/140602655
- Room schema location annotation processor argument throws off cacheability: https://issuetracker.google.com/issues/132245929

I believe the MergeJavaResourcesTask fix also reduces the impact of https://issuetracker.google.com/issues/142890134 since it avoids regenerating the merged resources in many cases.

This also drops fixes for older versions of Android as these are not widely used and the effort/complexity involved to support the binary incompatibilities is probably not worthwhile.


src/main/groovy/org/gradle/android/AndroidCacheFixPlugin.groovy
```diff
@@ -90,8 +95,12 @@ class AndroidCacheFixPlugin implements Plugin<Project> {
     }
 
     /**
-     * Fix {@link org.gradle.api.tasks.compile.CompileOptions#getBootClasspath()} introducing relocatability problems for {@link AndroidJavaCompile}.
+     * The following are kept for reference purposes.
```
lptr, line 98: I think Git history should be enough to keep these around.

src/main/groovy/org/gradle/android/workarounds/MergeJavaResourcesWorkaround.groovy
```diff
@@ -0,0 +1,20 @@
+package org.gradle.android.workarounds
+
+import com.android.build.gradle.internal.tasks.MergeJavaResourceTask
+import org.gradle.android.AndroidIssue
+import org.gradle.api.Task
+import org.gradle.api.file.FileCollection
+
+import java.lang.reflect.Method
+
+@AndroidIssue(introducedIn = "3.5.0", fixedIn = ["4.1.0"], link = "https://issuetracker.google.com/issues/140602655")
```
lptr, line 10: Would be nice to copy the description from the issue here as Javadoc.

src/main/groovy/org/gradle/android/workarounds/AbstractAbsolutePathWorkaround.groovy
```diff
@@ -0,0 +1,35 @@
...
+                    task.inputs.files(originalPropertyValue)
+                        .withPathSensitivity(PathSensitivity.RELATIVE)
+                        .withPropertyName("${propertyName}.workaround")
+                        .skipWhenEmpty(true)
+                }
+            }
+            // Set the task property back to its original value
+            task.doFirst {
+                setPropertyValue(task, originalPropertyValue)
+            }
```
lptr, line 32: Does this work with Gradle 6.4+? IIUC the instant execution team is adding some safe-guards against tasks changing their input properties, and with build cache we are going to add more soon.

build.gradle
```diff
@@ -22,29 +22,33 @@ version = ["git", "describe", "--match", "[0-9]*", "--dirty"].execute().text.tri
...
-    "3.0.1": ["4.1", "4.2.1", "4.3.1", "4.4.1", "4.5.1", "4.6", "4.7", "4.8.1", "4.9"],
-    "3.0.0": ["4.1"],
+    "4.0.0-beta04": ["6.1.1", "6.2.2", "6.3"],
+    "3.6.2": ["5.6.4", "6.0.1", "6.1.1", "6.2.2", "6.3"],
+    "3.6.1": ["5.6.4"],
+    "3.6.0": ["5.6.4"],
+    "3.5.3": ["5.4.1", "5.5.1", "5.6.4", "6.0.1", "6.1.1", "6.2.2", "6.3"],
+    "3.5.2": ["5.4.1"],
+    "3.5.1": ["5.4.1"],
+    "3.5.0": ["5.4.1"]
```
lptr, line 32: It would be good to test with some 6.4 nightly, too.

src/test/groovy/org/gradle/android/RelocationTest.groovy
```diff
@@ -1,286 +0,0 @@
...
-import spock.lang.Unroll
-
-import static org.gradle.android.Versions.android
-import static org.gradle.android.Versions.gradle
-import static org.gradle.testkit.runner.TaskOutcome.FROM_CACHE
-import static org.gradle.testkit.runner.TaskOutcome.NO_SOURCE
-import static org.gradle.testkit.runner.TaskOutcome.SUCCESS
-import static org.gradle.testkit.runner.TaskOutcome.UP_TO_DATE
-
-class RelocationTest extends AbstractTest {
```
lptr, line 17: Why remove this test? This used to check if the workarounds actually worked.

lptr: I would add one more thing to any test we end up having: explicitly check for deprecation warnings emitted. Just to make sure we don't end up using some deprecated APIs unintentionally.

src/test/groovy/org/gradle/android/SimpleAndroidApp.groovy
```diff
@@ -103,51 +104,108 @@ class SimpleAndroidApp {
...
+     * project compiles RenderScript with support mode enabled.  The property can be directly changed on some versions
+     * of android but not others.  Building native libraries as part of the project has external infrastructure
+     * requirements in that versions of cmake and ndk must be installed on the system or the test will fail. Adding
+     * RenderScript compilation to the project is the simplest way to trigger the problem, but support mode does not
+     * work reliably on all versions of android with MacOS Catalina (because of 32-bit support).  Thus, the following
+     * hack was born.  We enable support mode in order to trick AGP into setting up the affected property on
+     * MergeNativeLibsTask, then we disable it before execution (in order to avoid runtime errors related to 32-bit
+     * support) and we provide a dummy library for the downstream task to "merge".  This reliably triggers the
+     * cache relocation problem on all android versions.
+     */
```
lptr, line 168: Oh, that must have been painful! :o

src/test/groovy/org/gradle/android/CrossVersionOutcomeAndRelocationTest.groovy
```diff
@@ -0,0 +1,421 @@
...
+                remainingTasks.remove(taskName)
+            }
+            if (!remainingTasks.empty) {
+                remainingTasks.each { taskName ->
+                    def taskOutcome = result.task(taskName)?.outcome
+                    println "> Task '$taskName' executed with outcome $taskOutcome but was not expected"
+                    allMatched = false
+                }
+            }
+            return allMatched
```
lptr, line 94: I like how this is better readable now. What I'd add on top is an explicit check for deprecation warnings emitted. Just to make sure we don't end up using some deprecated APIs unintentionally.

src/main/groovy/org/gradle/android/workarounds/MergeNativeLibsWorkaround.groovy
```diff
@@ -0,0 +1,34 @@
...
+
+import java.lang.reflect.Field
+
+/**
+ * In 3.5.x, projectNativeLibs is not directly backed by state in the Task object.  Instead, it is effectively a
+ * method that calculates a new FileCollection from variantScope every time it is called.  So we only
+ * support this workaround for 3.6.x and 4.0.x.
+ */
+@AndroidIssue(introducedIn = "3.6.0", fixedIn = ["4.1.0"], link = "https://issuetracker.google.com/issues/140602655")
+class MergeNativeLibsWorkaround extends AbstractAbsolutePathWorkaround {
```
CristianGM, line 16: I think the real issue with MergeNativeLibsWorkaround is that it shouldn't be cacheable at all, it produces a big output that is more expensive to download than to execute (even when you have not a single NativeLib on the project).
Scans say you avoid some time using the build cache but it's only because the time used for the first build includes gzip and upload.

lptr: Please mention these concerns on the Android plugin ticket: https://issuetracker.google.com/issues/140602655

src/test/groovy/org/gradle/android/SimpleAndroidApp.groovy
```diff
@@ -103,51 +104,108 @@ class SimpleAndroidApp {
...
             repositories {
                 google()
                 jcenter()
             }
 
+            dependencies {
+                def room_version = "2.2.5"
+
+                implementation "androidx.room:room-runtime:\$room_version"
+                annotationProcessor "androidx.room:room-compiler:\$room_version" // For Kotlin use kapt instead of annotationProcessor
```
devisnik, line 129: Why don't we use `kapt` here?

devisnik: IIRC, when not using `kapt` or running with `kapt.use.worker.api=false`, setting a relative path for `room.schemaLocation` within the annotationProcessorOptions would work as well.

src/main/groovy/org/gradle/android/AndroidCacheFixPlugin.groovy
ghale: Yeah, if you know to look for it.  I left these around more for someone new to the plugin to have some examples of how to solve certain problems (beyond the few examples in the existing workarounds).  I don't feel strongly about it, and admittedly, it might be a bit much in terms of noise, but they're at least easier to discover this way.

src/main/groovy/org/gradle/android/workarounds/AbstractAbsolutePathWorkaround.groovy
ghale: Seems to work with the latest 6.4 nightly, but I wouldn't be surprised if this was locked down at some point for good reason.  It makes me feel dirty.

build.gradle
ghale: I've added a 6.4 nightly in here, too.

src/main/groovy/org/gradle/android/workarounds/RoomSchemaLocationWorkaround.groovy
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * Changes annotation processor arguments to set the room.schemaLocation via a CommandLineArgsProcessor
+ * which accurately sets it as an output of the task.  Note that it is common for people to set the value
+ * using the defaultConfig in the android extension, which causes every compile task to use the same
+ * output directory, causing overlapping outputs and breaking cacheability again.  To avoid this, we
+ * change the location for each task to treat the specified directory as a "base" directory, but derive
+ * the task-specific directory off of the task path, ensuring uniqueness across compile tasks.
+ */
+@AndroidIssue(introducedIn = "3.5.0", fixedIn = [], link = "https://issuetracker.google.com/issues/132245929")
+class RoomSchemaLocationWorkaround implements Workaround {
```
runningcode, line 17: This looks great. I was curious, while we're fixing this cache issue, does it also make sense to mark this schemas directory as an input to the test and androidTest tasks? Otherwise changes to the schema might not re-run the test if the tests are cached.

src/test/groovy/org/gradle/android/SimpleAndroidApp.groovy
ghale: I've added some changes to the room location workaround to handle things for kapt tasks as well (and added tests for it).  Note that there's only so much we can do because of the way kapt handles annotation processor arguments.  For instance, it automatically adds each individual ap argument as a separate input on the task, which means there is no opportunity for us to later prevent that argument from being considered as an input (and mark it as an output directory instead).  So, we change the argument to be relative, which handles the cacheability problem at least.  Right now, it's the best/only option available to us.

src/main/groovy/org/gradle/android/workarounds/RoomSchemaLocationWorkaround.groovy
ghale: That's a good suggestion.  I'll check it out and possibly implement that in a follow up change.  For now, I want to get an initial version of this out and see how well the existing fixes work.

PR #81, opened 2020-05-04
Fix deprecation warnings.

### LINKED ISSUE #80: RoomSchemaLocationWorkaround added new deprecations
With some of the changes in 1.0.2, we added some unnecessary deprecations:

image

And

image

Neither of these are necessary and should be removed.

runningcode:
This sets the `mergeRoomSchemaLocations` as an input annotated property.

This also set the duplicate copy strategy to include duplicates.

Fixes #80


src/main/groovy/org/gradle/android/workarounds/RoomSchemaLocationWorkaround.groovy
```diff
@@ -199,6 +201,7 @@ class RoomSchemaLocationWorkaround implements Workaround {
      * of those tasks if only a single variant is being assembled.
      */
     static class RoomSchemaLocationMergeTask extends DefaultTask {
+        @Input
```
ghale, line 204: I think this should be `@Internal`.  I explicitly did not want this task to be incremental (and I'm not sure `@Input` would work in an expected way here, either).

PR #82, opened 2020-05-05
Make room schema workaround work with KaptWithKotlincTask

ghale:
Previously, the room schema workaround would not be applied when `kapt.use.worker.api=false` because this causes the kapt tasks to use a different type (`KaptWithKotlincTask`) which stores the annotation processor options differently and evaluates relative paths differently.

This fixes the workaround so that it will also work when this option is set.  It also checks the Kotlin Gradle plugin version and punts if a version < 1.3.70 is used as the annotation processor options are stored differently on earlier versions and it's more difficult/complex to handle those additional cases.  For now, we only support 1.3.70 and higher.

### DIFF
--- a/src/main/groovy/org/gradle/android/workarounds/RoomSchemaLocationWorkaround.groovy
+++ b/src/main/groovy/org/gradle/android/workarounds/RoomSchemaLocationWorkaround.groovy
@@ -30,7 +30,8 @@ import java.lang.reflect.Field
  * For kapt tasks, we cannot implement a CommandLineArgsProcessor, so we change the value to be relative
  * for the purposes of snapshotting inputs, then change it to the absolute path of a task-specific directory
  * right before the task executes.  We then merge it back to the originally specified directory using the
- * aforementioned merge task.
+ * aforementioned merge task.  Note that this workaround only works with Kotlin Gradle plugin 1.3.70 or
+ * higher.
  *
  */
 @AndroidIssue(introducedIn = "3.5.0", fixedIn = [], link = "https://issuetracker.google.com/issues/132245929")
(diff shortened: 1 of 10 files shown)


src/main/groovy/org/gradle/android/workarounds/RoomSchemaLocationWorkaround.groovy
```diff
@@ -127,40 +155,115 @@ class RoomSchemaLocationWorkaround implements Workaround {
...
-    private void configureSchemaLocationOutputs(Task task) {
-        def processorOptions = getCompilerPluginOptions(task)
-        processorOptions.each { option ->
-            if (option.key == ROOM_SCHEMA_LOCATION) {
-                def relativePath = option.value
-                def schemaDestinationDir = task.project.file(relativePath)
-                def taskSpecificSchemaDir = getTaskSpecificSchemaDir(task)
+    static def getEncodedCompilerPluginOptions(Task task) {
+        def pluginOptionsField = task.class.superclass.getDeclaredField("pluginOptions")
+        pluginOptionsField.setAccessible(true)
```
runningcode, line 160: is this required? i always that that groovy could access private fields without reflection.

src/main/groovy/org/gradle/android/workarounds/RoomSchemaLocationWorkaround.groovy
```diff
@@ -104,12 +112,32 @@ class RoomSchemaLocationWorkaround implements Workaround {
...
+                        }
+                    }
+                }
+
+                doFirst {
+                    setRoomSchemaLocationToTaskSpecificDirForKaptWithoutKotlinc(task)
+                }
+            }
+
+            project.tasks.withType(kaptWithKotlincTaskClass) { Task task ->
```
runningcode, line 127: can we lazily configure the task? or would this break the workarounds?
```suggestion
            project.tasks.withType(kaptWithKotlincTaskClass).configureEach { Task task ->
```

runningcode:
Nice fix to only look up the kotlin version once!

src/main/groovy/org/gradle/android/workarounds/RoomSchemaLocationWorkaround.groovy
```diff
@@ -127,40 +155,115 @@ class RoomSchemaLocationWorkaround implements Workaround {
...
     static Class<?> getKaptWithoutKotlincTaskClass() {
         return Class.forName("org.jetbrains.kotlin.gradle.internal.KaptWithoutKotlincTask")
     }
 
+    static Class<?> getKaptWithKotlincTaskClass() {
+        return Class.forName("org.jetbrains.kotlin.gradle.internal.KaptWithKotlincTask")
+    }
+
+    VersionNumber getKotlinVersion() {
+        def projectPropertiesStream = this.class.classLoader.getResourceAsStream("project.properties")
```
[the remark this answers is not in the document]
ghale: Good call - no reason why we can't get this once and store it - I've changed it to do so.  

Regarding api - there are some anonymous Kotlin functions on the Plugin object that the Kotlin plugin provides for this, but they're available only on the instantiated object.  In our case, we want to make this decision possibly _before_ the kotlin plugin has been applied (and a plugin object is available).  I don't see a reason why this stuff couldn't be more easily available - I'll submit a feature request for it.

src/main/groovy/org/gradle/android/workarounds/RoomSchemaLocationWorkaround.groovy
ghale: It's possible that groovy -> groovy might allow accessing private fields, but groovy -> java definitely won't work without explicitly making it accessible.

src/main/groovy/org/gradle/android/workarounds/RoomSchemaLocationWorkaround.groovy
ghale: Good catch!  This was just an oversight on my part.  Fixed.

PR #83, opened 2020-05-13
Typo on Applying the plugin

CristianGM:
The name of the plugin was wrong here

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ plugins {
 }
 
 subprojects {
-    apply plugin: 'org.gradle.android-cache-fix`
+    apply plugin: 'org.gradle.android.cache-fix`
 }
 ```
 


CristianGM:
@ghale I think the tests are "wrong" just changing a readme shouldn't make them fail 🤔

ghale:
Thanks, @CristianGM!  The failure looks like an issue with the infrastructure (ndk missing).  I wonder if something changed with the default docker image or something.  I'll see what I can do to make these tests more resilient to that sort of thing.  I've got a few more tweaks in the queue, so I'll merge your PR and get it published when I've got those ready and the testing issues ironed out.

CristianGM:
I'm most interested in #84 because we want to use this plugin and check the improvements.

PR #85, opened 2020-05-14
Task specific room schema dir is populated with existing schemas

ghale:
Sometimes other annotation processors use the room schema location annotation processor argument to do further processing of schemas and expect the existing schemas to also be present (not just the newly generated schemas).  This change allows the task specific directory to be seeded with the existing schemas before starting the compilation process.

### DIFF
--- a/src/main/groovy/org/gradle/android/workarounds/RoomSchemaLocationWorkaround.groovy
+++ b/src/main/groovy/org/gradle/android/workarounds/RoomSchemaLocationWorkaround.groovy
@@ -105,6 +105,11 @@ class RoomSchemaLocationWorkaround implements Workaround {
 
                 // Register the generated schemas to be merged back to the original specified schema directory
                 task.project.roomSchemaMergeLocations.registerMerge(schemaDestinationDir, taskSpecificSchemaDir)
+
+                // Seed the task-specific generated schema dir with the existing schemas
+                task.doFirst {
+                    copyExistingSchemasToTaskSpecificTmpDir(task, schemaDestinationDir, taskSpecificSchemaDir)
+                }
             }
         )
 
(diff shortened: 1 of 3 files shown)


src/test/groovy/org/gradle/android/RoomSchemaLocationWorkaroundTest.groovy
```diff
@@ -215,23 +215,34 @@ class RoomSchemaLocationWorkaroundTest extends AbstractTest {
 
     void assertKaptSchemaOutputsExist() {
         // Task specific schemas
-        assert file("app/build/roomSchemas/kaptDebugKotlin/org.gradle.android.example.app.AppDatabase/2.json").exists()
-        assert file("app/build/roomSchemas/kaptReleaseKotlin/org.gradle.android.example.app.AppDatabase/2.json").exists()
-        assert file("library/build/roomSchemas/kaptDebugKotlin/org.gradle.android.example.library.AppDatabase/2.json").exists()
-        assert file("library/build/roomSchemas/kaptReleaseKotlin/org.gradle.android.example.library.AppDatabase/2.json").exists()
+        assertKaptSchemaOutputsExistFor("debug")
```
runningcode, line 218: i wonder if we should add a test for flavors as well. (debug and release are build types)

`freeDebug`, `freeRelease`, `paidDebug`, `paidRelease`

At the end of the day, its the same thing though. The output directory is the same.

src/main/groovy/org/gradle/android/workarounds/RoomSchemaLocationWorkaround.groovy
```diff
@@ -208,20 +233,44 @@ class RoomSchemaLocationWorkaround implements Workaround {
         task.project.roomSchemaMergeLocations.registerMerge(schemaDestinationDir, taskSpecificSchemaDir)
     }
 
-    private static void setRoomSchemaLocationToTaskSpecificDirForKaptWithoutKotlinc(Task task) {
+    private static void copyExistingSchemasToTaskSpecificTmpDir(Task task, File existingSchemaDir, File taskSpecificTmpDir) {
+        // populate the task-specific tmp dir with any existing (non-generated) schemas
+        // this allows other annotation processors that might operate on these schemas
+        // to find them via the schema location argument
+        task.project.sync {
```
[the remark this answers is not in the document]
ghale: `copy {}` will copy from location A to location B and leave anything else in B that was not in location A.  `sync {}` will do the copy, but remove anything in location B that was not in location A.

src/test/groovy/org/gradle/android/RoomSchemaLocationWorkaroundTest.groovy
ghale: Yeah, I think eventually the test application should be more realistic in general.  Right now, it has some code, resources, and a room database and some other things, but it's all very simplistic.  We've run into at least one issue where it didn't catch a problem because it was too simplistic.  I think it might be worth it at some point to create a test app that is a little more like a "real" application.

PR #86, opened 2020-05-15
Fix syntax highlighting and improve Gradle snippet.

johnjohndoe:
- Replace wrong trailing backtick to have snippet copy & paste ready.
- Use double quotes everywhere for consistency.
- Use consistent indentation.


ghale:
Thanks, @johnjohndoe!

PR #88, opened 2020-05-15
Fix CI checking for additional logs.

runningcode:
In github actions the CI variable is set.


runningcode:
Closed since the build scan plugin is applied instead.

PR #94, opened 2020-06-08
Add early support for AGP 4.1

runningcode:
Also bump versions and add local.properties to gitignore.
The travis property is removed since the build scan plugin is now applied to the tests.


ghale:
Thanks @runningcode!

PR #98, opened 2020-06-15
Fix CompileLibraryResources cache miss.

### LINKED ISSUE #97: Fix Remote Cache miss with CompileLibraryResourcesTask
CompileLibraryResourcesTask.mergedLibraryResourcesDir uses absolute path as input.

https://issuetracker.google.com/issues/155218379

runningcode:
This is a regression from AGP 3.6.

Fixes #97

### DIFF
--- a/src/main/groovy/org/gradle/android/workarounds/CompileLibraryResourcesWorkaround.groovy
+++ b/src/main/groovy/org/gradle/android/workarounds/CompileLibraryResourcesWorkaround.groovy
@@ -1,0 +1,62 @@
+package org.gradle.android.workarounds
+
+import org.gradle.android.AndroidIssue
+import org.gradle.api.Project
+import org.gradle.api.Task
+import org.gradle.api.file.DirectoryProperty
+import org.gradle.api.tasks.PathSensitivity
+
+import java.lang.reflect.Field
+
+
+/**
+ * Fixes the cacheability issue with CompileLibraryResourcesWorkaround where the mergedLibraryResourcesDir field is
+ * treated as an input with absolute path sensitivity.
+ */
+@AndroidIssue(introducedIn = "4.0.0", fixedIn = [], link = "https://issuetracker.google.com/issues/155218379")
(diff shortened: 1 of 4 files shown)


src/main/groovy/org/gradle/android/workarounds/CompileLibraryResourcesWorkaround.groovy
```diff
@@ -0,0 +1,63 @@
...
+
+    @Override
+    void apply(WorkaroundContext context) {
+        Project project = context.project
+        project.tasks.withType(androidTaskClass).configureEach { Task task ->
+            DirectoryProperty originalPropertyValue
+            // Create a synthetic input with the original property value and RELATIVE path sensitivity
+            project.gradle.taskGraph.beforeTask {
+                if (it == task) {
+                    println("configuring workaround before $task")
```
ghale, line 39: we need to remove this

build.gradle
```diff
@@ -111,6 +111,7 @@ task install(type: Upload) {
 test {
     dependsOn install
     systemProperty "local.repo", localRepo.toURI()
+    maxParallelForks = 2
```
ghale, line 114: Did this actually improve anything for you?  I think I tried this out at one point and decided that with it running the android builds under the covers, it didn't actually speed things up (i.e. the bottleneck is IO and more parallelization in the tests didn't do anything to work around that).

src/main/groovy/org/gradle/android/workarounds/CompileLibraryResourcesWorkaround.groovy
runningcode: haha good catch.

build.gradle
runningcode: Yeah, you are right. There's almost no difference here in terms of performance of the tests. 
max parallel forks enabled:
https://scans.gradle.com/s/zmoo7agh4z23a/performance/build
max parallel forks disabled:
https://scans.gradle.com/s/ikgh3usjl3wwc/performance/build

Will remove.

PR #102, opened 2020-07-10
Fix applying fixes to newer versions.

### LINKED ISSUE #99: AGP 4.2 incompatibility
Tested with 4.2.0-alpha04

```
Caused by: java.lang.NullPointerException: Cannot invoke method setAccessible() on null object
	at org.gradle.android.workarounds.MergeJavaResourcesWorkaround.setPropertyValue(MergeJavaResourcesWorkaround.groovy:22)
	at org.gradle.android.workarounds.AbstractAbsolutePathWorkaround$_apply_closure1$_closure2.doCall(AbstractAbsolutePathWorkaround.groovy:27)
```

runningcode:
There was a small bug in the logic which was causing the fixes to be applied to versions that are newer than the fixed version.

Fixes #99

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -22,8 +22,8 @@ version = ["git", "describe", "--match", "[0-9]*", "--dirty"].execute().text.tri
 
 // Maps supported Android plugin versions to the versions of Gradle that support it
 def supportedVersions = [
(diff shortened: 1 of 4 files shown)


ghale:
Wow.  I'm shocked we didn't notice this before now.

PR #116, opened 2020-08-11
Make DexFileDependencies cacheable

### LINKED ISSUE #111: Make DexFileDependenciesTask cacheable.
It will run by default starting in AGP 4.2.0-alpha05 and is likely running by default in other builds. We should make it cacheable.

Source: https://android.googlesource.com/platform/tools/base/+/studio-master-dev/build-system/gradle-core/src/main/java/com/android/build/gradle/internal/tasks/DexFileDependenciesTask.kt

Issue tracked here: https://issuetracker.google.com/160138798

runningcode:
Fixes #111

### DIFF
--- a/src/test/groovy/org/gradle/android/CrossVersionOutcomeAndRelocationTest.groovy
+++ b/src/test/groovy/org/gradle/android/CrossVersionOutcomeAndRelocationTest.groovy
@@ -477,8 +477,7 @@ class CrossVersionOutcomeAndRelocationTest extends AbstractTest {
     }
 
     static void android42xOrHigherExpectations(ExpectedOutcomeBuilder builder) {
-        // Should be made cacheable: https://issuetracker.google.com/160138798
-        builder.expect(':app:desugarDebugFileDependencies', SUCCESS)
-        builder.expect(':app:desugarReleaseFileDependencies', SUCCESS)
+        builder.expect(':app:desugarDebugFileDependencies', FROM_CACHE)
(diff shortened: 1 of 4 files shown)


src/test/groovy/org/gradle/android/CrossVersionOutcomeAndRelocationTest.groovy
```diff
@@ -477,8 +477,7 @@ class CrossVersionOutcomeAndRelocationTest extends AbstractTest {
     }
 
     static void android42xOrHigherExpectations(ExpectedOutcomeBuilder builder) {
-        // Should be made cacheable: https://issuetracker.google.com/160138798
-        builder.expect(':app:desugarDebugFileDependencies', SUCCESS)
-        builder.expect(':app:desugarReleaseFileDependencies', SUCCESS)
+        builder.expect(':app:desugarDebugFileDependencies', FROM_CACHE)
+        builder.expect(':app:desugarReleaseFileDependencies', FROM_CACHE)
```
ghale, line 481: Does this task only show up in AGP 4.2?  If so, why do we have the workaround being applied to earlier versions?  Or is it that the task is not triggered to execute in earlier versions?

runningcode: Yeah, good question. This task is not enabled by default in earlier versions but it still exists. Google has decided to enable it by default starting with 4.2.

PR #119, opened 2020-09-03
Updates for 4.2 alpha

runningcode:


src/main/groovy/org/gradle/android/workarounds/NewCompileLibraryResourcesWorkaround.groovy
```diff
@@ -0,0 +1,64 @@
...
+
+
+/**
+ * Fixes the cacheability issue with CompileLibraryResourcesWorkaround where the inputDirectories field is
+ * treated as an input with absolute path sensitivity.
+ * This is the same as the {@link CompileLibraryResourcesWorkaround} but the field was renamed with a new type
+ * in 4.2.0-alpha09.
+ */
+@AndroidIssue(introducedIn = "4.2.0-alpha09", fixedIn = [], link = "https://issuetracker.google.com/issues/155218379")
+class NewCompileLibraryResourcesWorkaround implements Workaround {
```
ghale, line 19: Is the only difference between this and `CompileLibraryResourcesWorkaround` the property name?  This feels like a lot of duplication.  Would it be better to just subclass the two implementations and have them return a different property name?  (maybe `CompileLibraryResourcesWorkaround_4_0` and `CompileLibraryResourcesWorkaround_4_2`?). Or even just keep one class and return a different property name depending on the Android version?

ghale (changes requested):
I would prefer to see this with less duplication, but I don't feel strong about it if there's something I'm missing.  Let me know what you think...

src/main/groovy/org/gradle/android/workarounds/NewCompileLibraryResourcesWorkaround.groovy
runningcode: Yeah, there's a lot of duplication there.
The thing is that it isn't a simple property rename, but also the type of the property has changed from a `DirectoryProperty` to a `FileCollection` which means the logic is also slightly changed here.
I'm not sure a base class or a version check it would make it simpler. What are your thoughts?

ghale: Understood.  I didn't realize there were some further differences there.  I think it's fine for now.

ghale (approved):

PR #121, opened 2020-09-10
Avoid configuring JavaCompile tasks eagerly.

### LINKED ISSUE #120: JavaCompile tasks are created during configuration
I noticed that after applying plugin all `JavaCompile` tasks are created during configuration. In the build scan report for `gradle help --scan` such tasks appear under "Created during configuration" category.

Is that known/deliberate behavior?

runningcode:
Fixes #120

### DIFF
--- a/src/test/groovy/org/gradle/android/TaskAvoidanceTest.groovy
+++ b/src/test/groovy/org/gradle/android/TaskAvoidanceTest.groovy
@@ -6,25 +6,20 @@ import spock.lang.Unroll
 
 class TaskAvoidanceTest extends AbstractTest {
     @Unroll
-    def "tasks are avoided with #gradleVersion and Android plugin #androidVersion"() {
+    def "Source Tasks are avoided with #gradleVersion and Android plugin #androidVersion"() {
         assert gradleVersion instanceof GradleVersion
         assert androidVersion instanceof VersionNumber
 
-        println "> Using Android plugin $androidVersion"
-        println "> Running with $gradleVersion"
-
+        given:
         SimpleAndroidApp.builder(temporaryFolder.root, cacheDir)
                 .withAndroidVersion(androidVersion)
-        .withKotlinDisabled()
+                .withKotlinDisabled()
                 .build()
                 .writeProject()
 
-        def originalSettings = file('build.gradle').text
-        file('build.gradle').text = """
-            ${originalSettings}
-
+        file('build.gradle') << """
             allprojects {
-              tasks.configureEach {
+              tasks.withType(SourceTask).configureEach {
                 println "configuring \$it"
               }
             }
@@ -33,21 +28,11 @@ class TaskAvoidanceTest extends AbstractTest {
         when:
         def result = withGradleVersion(gradleVersion.version)
                 .withProjectDir(temporaryFolder.root)
-                .withArguments("help")
+                .withArguments('help')
                 .build()
 
         then:
-        result.output.contains("configuring task ':help'")
-        !result.output.contains("configuring task ':app:compileDebugJavaWithJavac'")
-        !result.output.contains("configuring task ':app:compileDebugAndroidTestJavaWithJavac'")
-        !result.output.contains("configuring task ':app:compileDebugUnitTestJavaWithJavac'")
-        !result.output.contains("configuring task ':app:compileReleaseJavaWithJavac'")
-        !result.output.contains("configuring task ':app:compileReleaseUnitTestJavaWithJavac'")
-        !result.output.contains("configuring task ':library:compileDebugJavaWithJavac'")
-        !result.output.contains("configuring task ':library:compileDebugAndroidTestJavaWithJavac'")
-        !result.output.contains("configuring task ':library:compileDebugUnitTestJavaWithJavac'")
-        !result.output.contains("configuring task ':library:compileReleaseJavaWithJavac'")
-        !result.output.contains("configuring task ':library:compileReleaseUnitTestJavaWithJavac'")
+        !result.output.contains("configuring")
 
         where:
         [androidVersion, gradleVersion] << Versions.SUPPORTED_VERSIONS_MATRIX.entries().collect { [it.key, it.value] }
--- a/src/test/groovy/org/gradle/android/TaskAvoidanceTest.groovy
+++ b/src/test/groovy/org/gradle/android/TaskAvoidanceTest.groovy
@@ -1,0 +1,55 @@
+package org.gradle.android
+
+import org.gradle.util.GradleVersion
+import org.gradle.util.VersionNumber
+import spock.lang.Unroll
+
+class TaskAvoidanceTest extends AbstractTest {
+    @Unroll
+    def "tasks are avoided with #gradleVersion and Android plugin #androidVersion"() {
+        assert gradleVersion instanceof GradleVersion
+        assert androidVersion instanceof VersionNumber
+
+        println "> Using Android plugin $androidVersion"
+        println "> Running with $gradleVersion"
+
+        SimpleAndroidApp.builder(temporaryFolder.root, cacheDir)
+                .withAndroidVersion(androidVersion)
+        .withKotlinDisabled()
+                .build()
+                .writeProject()
+
+        def originalSettings = file('build.gradle').text
+        file('build.gradle').text = """
+            ${originalSettings}
+
+            allprojects {
+              tasks.configureEach {
+                println "configuring \$it"
+              }
+            }
+        """
+
+        when:
+        def result = withGradleVersion(gradleVersion.version)
+                .withProjectDir(temporaryFolder.root)
+                .withArguments("help")
+                .build()
+
+        then:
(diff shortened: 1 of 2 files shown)


src/test/groovy/org/gradle/android/TaskAvoidanceTest.groovy
```diff
@@ -0,0 +1,55 @@
...
+import spock.lang.Unroll
+
+class TaskAvoidanceTest extends AbstractTest {
+    @Unroll
+    def "tasks are avoided with #gradleVersion and Android plugin #androidVersion"() {
+        assert gradleVersion instanceof GradleVersion
+        assert androidVersion instanceof VersionNumber
+
+        println "> Using Android plugin $androidVersion"
+        println "> Running with $gradleVersion"
```
etiennestuder, line 14: Use same order in asserts as in printlns

etiennestuder: Why do we even have these asserts and printlns?

src/test/groovy/org/gradle/android/TaskAvoidanceTest.groovy
```diff
@@ -0,0 +1,55 @@
...
+
+        SimpleAndroidApp.builder(temporaryFolder.root, cacheDir)
+                .withAndroidVersion(androidVersion)
+        .withKotlinDisabled()
```
etiennestuder, line 18: Polish indenting

src/test/groovy/org/gradle/android/TaskAvoidanceTest.groovy
```diff
@@ -0,0 +1,55 @@
...
+class TaskAvoidanceTest extends AbstractTest {
+    @Unroll
+    def "tasks are avoided with #gradleVersion and Android plugin #androidVersion"() {
+        assert gradleVersion instanceof GradleVersion
+        assert androidVersion instanceof VersionNumber
+
+        println "> Using Android plugin $androidVersion"
+        println "> Running with $gradleVersion"
+
+        SimpleAndroidApp.builder(temporaryFolder.root, cacheDir)
```
etiennestuder, line 16: add `given:`

src/test/groovy/org/gradle/android/TaskAvoidanceTest.groovy
```diff
@@ -0,0 +1,55 @@
...
+        def originalSettings = file('build.gradle').text
+        file('build.gradle').text = """
+            ${originalSettings}
+
+            allprojects {
+              tasks.configureEach {
+                println "configuring \$it"
+              }
+            }
+        """
```
etiennestuder, line 31: I believe this can be simplified with `file('build.gradle') << " .... "

src/test/groovy/org/gradle/android/TaskAvoidanceTest.groovy
```diff
@@ -0,0 +1,55 @@
...
+
+        when:
+        def result = withGradleVersion(gradleVersion.version)
+                .withProjectDir(temporaryFolder.root)
+                .withArguments("help")
```
etiennestuder, line 36: single quotes are sufficient

src/test/groovy/org/gradle/android/TaskAvoidanceTest.groovy
```diff
@@ -0,0 +1,55 @@
...
+        .withKotlinDisabled()
+                .build()
+                .writeProject()
+
+        def originalSettings = file('build.gradle').text
+        file('build.gradle').text = """
+            ${originalSettings}
+
+            allprojects {
+              tasks.configureEach {
```
etiennestuder, line 27: I would only do these for type `AbstractCompile` and then in the assertion, ensure that there is no output that contains "configuring...".

src/test/groovy/org/gradle/android/TaskAvoidanceTest.groovy
```diff
@@ -0,0 +1,55 @@
+package org.gradle.android
+
+import org.gradle.util.GradleVersion
+import org.gradle.util.VersionNumber
+import spock.lang.Unroll
+
+class TaskAvoidanceTest extends AbstractTest {
+    @Unroll
+    def "tasks are avoided with #gradleVersion and Android plugin #androidVersion"() {
```
etiennestuder, line 9: adjust name such that it reflects that we test for avoidance of compile tasks

src/test/groovy/org/gradle/android/TaskAvoidanceTest.groovy
```diff
@@ -0,0 +1,55 @@
...
+        !result.output.contains("configuring task ':app:compileDebugJavaWithJavac'")
+        !result.output.contains("configuring task ':app:compileDebugAndroidTestJavaWithJavac'")
+        !result.output.contains("configuring task ':app:compileDebugUnitTestJavaWithJavac'")
+        !result.output.contains("configuring task ':app:compileReleaseJavaWithJavac'")
+        !result.output.contains("configuring task ':app:compileReleaseUnitTestJavaWithJavac'")
+        !result.output.contains("configuring task ':library:compileDebugJavaWithJavac'")
+        !result.output.contains("configuring task ':library:compileDebugAndroidTestJavaWithJavac'")
+        !result.output.contains("configuring task ':library:compileDebugUnitTestJavaWithJavac'")
+        !result.output.contains("configuring task ':library:compileReleaseJavaWithJavac'")
+        !result.output.contains("configuring task ':library:compileReleaseUnitTestJavaWithJavac'")
```
etiennestuder, line 50: See comment above. You can then get rid of all these hard-coded task names.

etiennestuder (changes requested):

src/test/groovy/org/gradle/android/TaskAvoidanceTest.groovy
runningcode: Good point, but maybe it's good to keep this as is in order to test that the Android Gradle Plugin does not regresses in any way in terms of task configuration avoidance. One would argue that we shouldn't be testing this for Google, but that's basically what we do with the `CrossVersionOutcomeAndRelocationTest` by testing all the tasks in the Android Gradle Plugin.

src/test/groovy/org/gradle/android/TaskAvoidanceTest.groovy
runningcode: I wonder why codenarc - which was executed on CI on the test sources - did not cache this: https://scans.gradle.com/s/bkrkyiqc6bcrc/timeline?details=mx7qk7s42pldi&hideTimeline
Will fix.

src/test/groovy/org/gradle/android/TaskAvoidanceTest.groovy
runningcode: Not sure, was copypasta from here: https://github.com/gradle/android-cache-fix-gradle-plugin/blob/master/src/test/groovy/org/gradle/android/CrossVersionOutcomeAndRelocationTest.groovy#L20

The test method name should have the versions in it. Will remove

src/test/groovy/org/gradle/android/TaskAvoidanceTest.groovy
runningcode: nice groovy magic

src/test/groovy/org/gradle/android/TaskAvoidanceTest.groovy
runningcode: Ah, I see what you mean though after reading the comment below. The test is more thorogh. Will change.

src/test/groovy/org/gradle/android/TaskAvoidanceTest.groovy
```diff
@@ -0,0 +1,40 @@
+package org.gradle.android
+
+import org.gradle.util.GradleVersion
+import org.gradle.util.VersionNumber
+import spock.lang.Unroll
+
+class TaskAvoidanceTest extends AbstractTest {
+    @Unroll
+    def "Source Tasks are avoided with #gradleVersion and Android plugin #androidVersion"() {
+        assert gradleVersion instanceof GradleVersion
```
runningcode, line 10: After removing these, I learned that this helps with Intellij type inference. I added them back so that Intellij can properly show errors when these are passed to the functions incorrectly.

etiennestuder: Seems a bit contradictive to me to use a dyanmic language and then optimize the code for type inference by the IDE.

runningcode: Yeah makes sense. It was copied from here: https://github.com/gradle/android-cache-fix-gradle-plugin/blob/master/src/test/groovy/org/gradle/android/CrossVersionOutcomeAndRelocationTest.groovy#L20

Removed.

etiennestuder (approved):

etiennestuder:
@runningcode thanks for the polishing. good to merge.

Issue #101, opened 2020-07-10
Support Configuration Caching


runningcode:
Starting this issue to strategize about this:

Are we going to need to compile against Gradle 6.0 to take advantage of newer APIs?
Should we drop support for AGP 3.5? No new cache fixes against AGP 3.5 are likely going to be developed, maybe we can just suggest AGP 3.5 users stay with version 1.0.7?

Are we going to be able to do all the funny business we do with task manipulation with configuration caching?

ZacSweers:
```
org.gradle.api.InvalidUserCodeException: Read system property 'org.gradle.android.cache-fix.ignoreVersionCheck'
	at org.gradle.instantexecution.SystemPropertyAccessListener.systemPropertyQueried(SystemPropertyAccessListener.kt:92)
	at org.gradle.internal.classpath.Instrumented.getBoolean(Instrumented.java:133)
	at org.gradle.android.AndroidCacheFixPlugin.isSupportedAndroidVersion(AndroidCacheFixPlugin.groovy:53)
	at org.gradle.android.AndroidCacheFixPlugin.apply(AndroidCacheFixPlugin.groovy:63)
	at org.gradle.android.AndroidCacheFixPlugin.apply(AndroidCacheFixPlugin.groovy)
```

```
org.gradle.api.InvalidUserCodeException: Read system property 'org.gradle.android.cache-fix.ignoreVersionCheck'
	at org.gradle.instantexecution.SystemPropertyAccessListener.systemPropertyQueried(SystemPropertyAccessListener.kt:92)
	at org.gradle.internal.classpath.Instrumented.getBoolean(Instrumented.java:133)
	at org.gradle.android.AndroidCacheFixPlugin.isSupportedAndroidVersion(AndroidCacheFixPlugin.groovy:53)
	at org.gradle.android.AndroidCacheFixPlugin.apply(AndroidCacheFixPlugin.groovy:63)
	at org.gradle.android.AndroidCacheFixPlugin.apply(AndroidCacheFixPlugin.groovy)
```

ZacSweers:
```
at org.gradle.android.workarounds.CompilerArgsProcessor$_ensureApplied_closure1.doCall(CompilerArgsProcessor.groovy:44)
```

ZacSweers:
```
at org.gradle.android.workarounds.RoomSchemaLocationWorkaround$_apply_closure6$_closure17.doCall(RoomSchemaLocationWorkaround.groovy:124)
[...]

PR #131, opened 2021-01-09
Use systemProperty on Gradle 6.1 or newer

ZacSweers:
Relates to #101. There might be others but this was the only one in our project's report


ZacSweers:
Not sure why DCO doesn't like my commit even though it's signed 
image

ZacSweers:
CC @runningcode

ghale (approved):
Thanks, Zach, looks good.

PR #132, opened 2021-01-10
Add missing words in README

LouisCAD:
Otherwise, it reads like it's a version of the Android OS, which is not.


LouisCAD:
Screw you DCO bot, I'm not spending my time cloning the entire repo to edit just one file.

ghale:
Thanks, @LouisCAD!

Issue #135, opened 2021-01-29
Cannot get property 'temporarySchemaLocationDir' on null object


kozaxinan:
When project applies gradle cache fix 2.0.0 on a large project with room extension, build fails on a module that is not even android.

```
Execution failed for task ':some-module:kaptKotlin'.
> Cannot get property 'temporarySchemaLocationDir' on null object
```

`some-module` module is pure kotlin module with kapt plugin.

```
* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':base-networking:kaptKotlin'.
	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.lambda$executeIfValid$3(ExecuteActionsTaskExecuter.java:186)
	at org.gradle.internal.Try$Failure.ifSuccessfulOrElse(Try.java:268)
	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeIfValid(ExecuteActionsTaskExecuter.java:184)
	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:173)
	at org.gradle.api.internal.tasks.execution.CleanupStaleOutputsExecuter.execute(CleanupStaleOutputsExecuter.java:109)
	at org.gradle.api.internal.tasks.execution.FinalizePropertiesTaskExecuter.execute(FinalizePropertiesTaskExecuter.java:46)
	at org.gradle.api.internal.tasks.execution.ResolveTaskExecutionModeExecuter.execute(ResolveTaskExecutionModeExecuter.java:62)
	at org.gradle.api.internal.tasks.execution.SkipTaskWithNoActionsExecuter.execute(SkipTaskWithNoActionsExecuter.java:57)
	at org.gradle.api.internal.tasks.execution.SkipOnlyIfTaskExecuter.execute(SkipOnlyIfTaskExecuter.java:56)
	at org.gradle.api.internal.tasks.execution.CatchExceptionTaskExecuter.execute(CatchExceptionTaskExecuter.java:36)
	at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.executeTask(EventFiringTaskExecuter.java:77)
	at org.gradle.api.internal.tasks.execution.EventFiringTaskExecuter$1.call(EventFiringTaskExecuter.java:55)
[...]

PR #136, opened 2021-02-01
Only apply Room configuration to main compile tasks

ghale:
Fixes the issue described in https://github.com/gradle/android-cache-fix-gradle-plugin/issues/135#issuecomment-770726740 where the Room workaround configuration is only partially be applied to the test compile tasks.  Configuration is now only applied to the main compile tasks.  Tests have been expanded to cover test tasks as well in order to catch these sorts of issues.

### DIFF
--- a/src/test/groovy/org/gradle/android/RoomSchemaLocationWorkaroundTest.groovy
+++ b/src/test/groovy/org/gradle/android/RoomSchemaLocationWorkaroundTest.groovy
@@ -6,8 +6,13 @@ import org.gradle.testkit.runner.TaskOutcome
 import org.gradle.util.VersionNumber
 import spock.lang.Unroll
 
+import static org.gradle.testkit.runner.TaskOutcome.FROM_CACHE
+import static org.gradle.testkit.runner.TaskOutcome.SUCCESS
+
 class RoomSchemaLocationWorkaroundTest extends AbstractTest {
-    private static final String[] CLEAN_BUILD = ["clean", "testDebug", "testRelease", "--build-cache", "--stacktrace"]
+    private static final String[] CLEAN_BUILD = ["clean", "testDebug", "testRelease", "assembleAndroidTest", "--build-cache", "--stacktrace"]
+    private static final List<String> ALL_PROJECTS = ["app", "library"]
+    private static final List<String> ALL_VARIANTS = ["debug", "release"]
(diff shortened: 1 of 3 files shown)


src/test/groovy/org/gradle/android/RoomSchemaLocationWorkaroundTest.groovy
```diff
@@ -7,6 +7,8 @@ import org.gradle.util.VersionNumber
 import spock.lang.Unroll
 
 class RoomSchemaLocationWorkaroundTest extends AbstractTest {
+    private static final String[] CLEAN_BUILD = ["clean", "testDebug", "testRelease", "--build-cache", "--stacktrace"]
```
runningcode, line 10: can you add `assembleAndroidTest` here? we should also make sure that the androidTest can compile since kapt also adds itself there.

src/test/groovy/org/gradle/android/RoomSchemaLocationWorkaroundTest.groovy
```diff
@@ -42,7 +44,7 @@ class RoomSchemaLocationWorkaroundTest extends AbstractTest {
         buildResult = withGradleVersion(Versions.latestGradleVersion().version)
             .forwardOutput()
             .withProjectDir(temporaryFolder.root)
-            .withArguments("clean", "assemble", "--build-cache", "--stacktrace")
+            .withArguments(CLEAN_BUILD)
             .build()
 
         then:
```
runningcode, line 50: should we also make sure that the unitTest tasks came from the cache? Meaning that there wasn't a cache miss caused by room? as well as the androidTestCompile when added.

src/test/groovy/org/gradle/android/RoomSchemaLocationWorkaroundTest.groovy
ghale: Good idea.  Done.

src/test/groovy/org/gradle/android/RoomSchemaLocationWorkaroundTest.groovy
ghale: I think so.  Done.

runningcode (approved):
Looking good!

PR #141, opened 2021-02-11
Disable caching for MergeNativeLibs.

runningcode:
This disables caching for the MergeNativeLibs task because it behaves like a copy task which do not benefit from caching.

This PR also removes the workaround for DexFileDependencies since this task is not cacheable by default.


src/main/groovy/org/gradle/android/workarounds/MergeNativeLibsWorkaround.groovy
```diff
@@ -2,37 +2,19 @@ package org.gradle.android.workarounds
...
-    String propertyName = "projectNativeLibs"
+import org.gradle.api.Project
 
+@AndroidIssue(introducedIn = "3.5.0", fixedIn = [], link = "https://issuetracker.google.com/issues/153088766")
+class MergeNativeLibsWorkaround implements Workaround {
     @Override
     void apply(WorkaroundContext context) {
-        super.apply(context)
+        context.project.tasks.withType(MergeNativeLibsTask).configureEach {
+            outputs.doNotCacheIf("Caching MergeNativeLibs adds overhead to the build and never saves time", {true })
```
ghale, line 12: This feels a little too strongly worded.  Perhaps "Caching MergeNativeLibs is unlikely to provide positive performance results".

etiennestuder:
I'd like for you and @ghale to decide.

ghale (changes requested):
Can you provide more color as to why `DexFileDependenciesWorkaround` is being removed?  Is this one where we've found that it doesn't provide significant gains?

src/main/groovy/org/gradle/android/workarounds/MergeNativeLibsWorkaround.groovy
runningcode: Updated!

ghale (approved):

PR #142, opened 2021-02-12
Improve the cache-fix plugin build

bigdaz:
- Enable caching
- Enable parallel test execution
- Use `maven-publish` instead of `Upload` task
- Use Java toolchains to compile with JDK8

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -44,8 +44,11 @@ dependencies {
     testRuntime 'org.objenesis:objenesis:3.1'
 }
 
-sourceCompatibility = "1.8"
-targetCompatibility = "1.8"
+java {
+    toolchain {
+        languageVersion = JavaLanguageVersion.of(8)
+    }
+}
 
 def generatedResources = "$buildDir/generated-resources/main"
 
@@ -103,8 +106,4 @@ tasks.withType(Test).configureEach {
         maxRetries = 3
         maxFailures = 20
     }
-}
-
-if (!JavaVersion.current().java8) {
-    throw new RuntimeException("Java 8 is required to execute the build")
 }
--- a/build.gradle
+++ b/build.gradle
@@ -1,10 +1,11 @@
 import groovy.json.JsonBuilder
 
 plugins {
-    id "com.gradle.plugin-publish" version "0.12.0"
     id 'groovy'
     id 'java-gradle-plugin'
+    id 'maven-publish'
     id 'codenarc'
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -67,48 +71,40 @@ sourceSets {
...
-    descriptorDestination = file("$buildDir/ivy/ivy.xml")
-    uploadDescriptor = true
-    configuration = configurations.runtime
 }
 
 tasks.withType(Test).configureEach {
-    dependsOn install
+    dependsOn publish
     systemProperty "local.repo", localRepo.toURI()
+    maxParallelForks = 4
```
ghale, line 105: Did you find that this actually improved anything?  When I've tried this before, with the android builds these tests execute, I found that it didn't improve the execution time at all, but it made my system nearly unusable while the tests were running.  It appeared that it was almost completely I/O bound, so adding more parallelism didn't buy us anything.

build.gradle
```diff
@@ -67,48 +71,40 @@ sourceSets {
...
-
+gradlePlugin {
     plugins {
         androidCacheFixPlugin {
             id = 'org.gradle.android.cache-fix'
             displayName = 'Gradle Android cache fix plugin'
-            tags = ['android', 'cache', 'fix']
-            version = project.version
+            description = 'Gradle plugin to fix Android caching problems'
+            implementationClass = 'org.gradle.android.AndroidCacheFixPlugin'
```
ghale, line 80: I think this means we can remove `src/main/resources/META-INF/gradle-plugins/org.gradle.android-cache-fix.properties` now, right?

ghale (changes requested):
Thanks for doing this.  Left a few minor comments, but otherwise it looks good.

build.gradle
bigdaz: I can't say I really noticed, but tests seemed super-slow without it. 
Looking at the CI job, it's probably not any faster. I'll remove this.

build.gradle
bigdaz: Good catch. Removed

bigdaz:
@ghale I'll merge this as-is and improve the test in a second pass. I'll also look at having a single `taskGraph.whenReady` hook, although I'm not overly concerned about leaving it with a separate hook per Workaround.

bigdaz:
@ghale Oops, that comment was meant for the other PR.

build.gradle
runningcode: Haha, yeah I did this too at some point :)

PR #143, opened 2021-02-12
Make plugin compatible with configuration-cache

bigdaz:


src/test/groovy/org/gradle/android/ConfigurationCachingTest.groovy
```diff
@@ -0,0 +1,21 @@
...
+                .writeProject()
+
+        when:
+        def result = withGradleVersion(Versions.latestGradleVersion().version)
+                .withProjectDir(temporaryFolder.root)
+                .withArguments('--configuration-cache', 'assembleDebug')
+                .build()
+
+        then:
+        !result.output.contains("problems were found storing the configuration cache")
```
ghale, line 19: Should we also run an additional execution and check that the configuration cache is reused on a follow up execution?  See https://github.com/gradle/test-retry-gradle-plugin/blob/master/plugin/src/test/groovy/org/gradle/testretry/ConfigCachingPluginFuncTest.groovy for an example.

src/test/groovy/org/gradle/android/ConfigurationCachingTest.groovy
```diff
@@ -0,0 +1,21 @@
+package org.gradle.android
+
+class ConfigurationCachingTest extends AbstractTest {
+    def "Can run with configuration cache"() {
+        given:
+        SimpleAndroidApp.builder(temporaryFolder.root, cacheDir)
+                .withAndroidVersion(Versions.latestAndroidVersion())
```
ghale, line 7: I wonder if we should iterate over the "config-cache-compatible" versions of AGP.  At least initially this would only be the latest 4.2 version, but as new versions are released, we might want to ensure that we aren't somehow breaking an older version.  Dunno - might be over-paranoid testing.  Your call.

ghale (approved):
Looks good!  I left some suggestions, but neither should block this, I think.

PR #151, opened 2021-03-03
Disable caching of MergeJavaResources, StripDebugSymbols

runningcode:


src/main/groovy/org/gradle/android/AndroidCacheFixPlugin.groovy
```diff
@@ -31,6 +31,7 @@ class AndroidCacheFixPlugin implements Plugin<Project> {
     private static final Logger LOGGER = LoggerFactory.getLogger(AndroidCacheFixPlugin)
 
     static final String IGNORE_VERSION_CHECK_PROPERTY = "org.gradle.android.cache-fix.ignoreVersionCheck"
+    static final String DO_NOT_DISABLE_CACHING_PROPERTY = "org.gradle.android.cache-fix.doNotDisableCaching"
```
ghale, line 34: I would not make this a global variable, but something specific to the `MergeJavaResourcesWorkaround`.  For example, `org.gradle.android.cache-fix.MergeJavaResources.enabled` which basically just enables you to toggle this workaround on or off.

ghale (changes requested):

ghale (approved):
I think this looks good.  It's still marked as a Draft, though - are you ready to merge this?

PR #152, opened 2021-03-04
Add stricter plugin validation.

runningcode:
Use task configuraiton avoidance.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -70,7 +70,7 @@ tasks.register('generateVersions') {
 
 sourceSets {
     main {
-        output.dir(generatedResources, builtBy: generateVersions)
+        output.dir(generatedResources, builtBy: tasks.named('generateVersions'))
     }
 }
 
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -129,7 +129,7 @@ tasks.withType(Test).configureEach {
     }
 }
 
-tasks.named("test") {
+tasks.named("test").configure {
```
ghale, line 132: This shouldn't actually have an effect.  `tasks.named(...) { }` and `tasks.named(...).configure { }` do the same thing.  See https://docs.gradle.org/current/dsl/org.gradle.api.tasks.TaskContainer.html#org.gradle.api.tasks.TaskContainer:named(java.lang.String,%20org.gradle.api.Action).

build.gradle
```diff
@@ -167,5 +169,10 @@ static def jdkVersionFor(String version) {
 }
 
 tasks.register("sanityCheck") {
-    dependsOn tasks.withType(CodeNarc)
+    dependsOn tasks.withType(CodeNarc), validatePlugins
```
ghale, line 172: good call

ghale (commented):
LGTM

PR #156, opened 2021-04-07
Update to latest version of 7.0.0-alpha and 4.2.0-rc01

ghale:
Three changes here:

- 7.0.0 breaks the MergeResources workaround.  The property we manipulate to make it use absolute path sensitivity is now made immutable once the task graph has been calculated.  This means that although we can zero out the existing input before fingerprinting starts, we can't set it back to its original value after fingerprinting is finished.  I don't see a way around this, so I think we'll just need to wait for a fix to the underlying issue.
- The `inputDirectories` property for `CompileLibraryResourcesTask` has been renamed again in 7.0.0.  Luckily, its type did not change, so we can just subclass the 4.0 version of the workaround with the only difference being the name of the property.
- In 7.0.0, the hack that was in place to force `MergeNativeLibs` to execute no longer works because the property it sets is no longer mutable at execution time.  Since we have changed this workaround to _not_ do any caching any more, it's not as important that the task execute, so I've just removed this hack (which was super questionable in the first place).  This resulted in some differing task outcomes across the various versions (sometimes SUCCESS, sometimes NO_SOURCE).

### DIFF
--- a/src/test/groovy/org/gradle/android/SimpleAndroidApp.groovy
+++ b/src/test/groovy/org/gradle/android/SimpleAndroidApp.groovy
@@ -225,8 +225,8 @@ class SimpleAndroidApp {
      * support) and we provide a dummy library for the downstream task to "merge".  This reliably triggers the
      * cache relocation problem on all android versions.
      */
-    private static String getRenderscriptConfiguration() {
-        return '''
+    private String getRenderscriptConfiguration() {
+        return androidVersion >= VersionNumber.parse('7.0.0-alpha12') ? '' : '''
             android {
                 defaultConfig {
                     renderscriptTargetApi 18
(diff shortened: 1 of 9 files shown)


src/test/groovy/org/gradle/android/SimpleAndroidApp.groovy
```diff
@@ -160,8 +160,6 @@ class SimpleAndroidApp {
             }
 
             ${roomExtensionIfEnabled}
-
-            ${renderscriptConfiguration}
```
runningcode, line 164: ah nice. i think this is a very rarely used feature. good to remove it.

runningcode (approved):
Looking good! thanks for taking care of this.

runningcode:
Fixes #155

PR #166, opened 2021-05-03
RoomSchemaLocationWorkaround: Change syntax in exception message

osipxd:
Replaced operator `=` with `.set` function call to make suggestion compatible with Kotlin DSL.


ghale:
Thanks @osipxd!

PR #168, opened 2021-05-06
Test latest kotlin version and oldest supported.

runningcode:
This adds tests for the newest and oldest supported kotlin versions in the `RoomSchemaLocationWorkaroundTest`.
The other relocation tests will default to the latest supported version.

### DIFF
--- a/src/test/groovy/org/gradle/android/TestVersions.groovy
+++ b/src/test/groovy/org/gradle/android/TestVersions.groovy
@@ -41,4 +41,21 @@ class TestVersions {
         def minorVersions = allCandidateTestVersions.keySet().collect { "${it.major}.${it.minor}" }
         return minorVersions.collect { getLatestVersionForAndroid(it) }
     }
+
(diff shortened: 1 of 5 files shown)


src/test/groovy/org/gradle/android/TestVersions.groovy
```diff
@@ -41,4 +41,21 @@ class TestVersions {
...
+
+    static VersionNumber oldestSupportedKotlinVersion() {
+        return VersionNumber.parse(supportedKotlinVersions.first())
+    }
+
+    static VersionNumber latestSupportedKotlinVersion() {
+        return VersionNumber.parse(supportedKotlinVersions.last())
+    }
+
+    static VersionNumber latestKotlinVersionForGradleVersion(GradleVersion gradleVersion) {
```
runningcode, line 55: @ghale I added this since KGP 1.5 doesn't support Gradle versions lower than 6.1

PR #174, opened 2021-07-02
Support AGP 4.2.2 and 7.0.0-beta05

runningcode:


ghale (changes requested):
This looks good, but it looks like there are some new tasks in the latest beta that need to be handled.  As far as I'm concerned, this is good to merge once that test is fixed.

zsperske:
Minor nitpick but should the title be updated to reflect "beta05" instead of "beta04"?

src/test/groovy/org/gradle/android/CrossVersionOutcomeAndRelocationTest.groovy
```diff
@@ -547,5 +547,15 @@ class CrossVersionOutcomeAndRelocationTest extends AbstractTest {
         builder.expect(':app:mergeReleaseResources', SUCCESS)
         builder.expect(':app:processDebugResources', SUCCESS)
         builder.expect(':app:processReleaseResources', SUCCESS)
+        // New tasks in 4.2.0-beta04
```
ghale, line 550: Are these actually new in 4.2.0?  If so, why are they in the "android7xOrHigher" list?

ghale (approved):

PR #180, opened 2021-07-30
Change CompileLibraryResourcesWorkaround to warn when experimental flags are not used

### LINKED ISSUE #170: Update in CompileLibraryResourcesTask implementation
According to https://issuetracker.google.com/issues/141301405#comment11 partial fix is available on mac and linux at (I assume) AGP 7.0-alpha16 with

> android.experimental.enableSourceSetPathsMap=true
android.experimental.cacheCompileLibResources=true

What is the path forward to agp-cache-fix-plugin implementation:
* wait until full fix is provided by Google
* continue to use agp-cache-fix-plugin overwrite
* check for properties and siable agp-cache-fix-plugin overwrite
?

ghale:
Fixes #170 

This disables our workaround for `CompileLIbraryResourcesTask` in AGP 7.0 and instead checks to see if the experimental support to fix this has been enabled.  If not, we simply print a warning telling the user that this support exists and how to enable it.


src/main/groovy/org/gradle/android/Warnings.groovy
```diff
@@ -0,0 +1,23 @@
+package org.gradle.android
+
+import org.gradle.android.workarounds.CompileLibraryResourcesWorkaround_7_0
+
+import java.util.concurrent.atomic.AtomicBoolean
+
+enum Warnings {
```
runningcode, line 7: Using the enum is a clever singleton I just realized!
Does this mean it will be warned once per Gradle daemon or once per build?

src/test/groovy/org/gradle/android/CrossVersionOutcomeAndRelocationTest.groovy
```diff
@@ -541,15 +541,18 @@ class CrossVersionOutcomeAndRelocationTest extends AbstractTest {
...
         builder.expect(':app:mergeDebugResources', SUCCESS)
         builder.expect(':app:mergeReleaseResources', SUCCESS)
-        builder.expect(':app:processDebugResources', SUCCESS)
-        builder.expect(':app:processReleaseResources', SUCCESS)
+        builder.expect(':app:processDebugResources', FROM_CACHE)
+        builder.expect(':app:processReleaseResources', FROM_CACHE)
         // New tasks in 7.0.0-beta04
         builder.expect(':library:javaPreCompileDebug', FROM_CACHE)
         builder.expect(':library:javaPreCompileRelease', FROM_CACHE)
+        builder.expect(':library:mapReleaseSourceSetPaths', SUCCESS)
```
runningcode, line 555: another one? :(

runningcode (approved):
Looks good to me. Thanks for investigating.

src/main/groovy/org/gradle/android/Warnings.groovy
ghale: That's a good question.  Turns out it's once per daemon due to classloader caching.  Perhaps we should reset this at the end of each build session.

PR #181, opened 2021-08-05
Fix issues with room schema merge

ghale:
This fixes an issue where an incremental build of a single variant with a Room schema change can result in the wrong schemas being merged to the target folder.  

It also changes the way we handle JavaCompile vs Kapt handling so that it's more resilient and less likely to result in incorrect schemas being merged.

### DIFF
--- a/src/main/groovy/org/gradle/android/workarounds/RoomSchemaLocationWorkaround.groovy
+++ b/src/main/groovy/org/gradle/android/workarounds/RoomSchemaLocationWorkaround.groovy
@@ -74,6 +74,9 @@ class RoomSchemaLocationWorkaround implements Workaround {
         // Grab fileOperations so we can do copy/sync operations
         def fileOperations = project.fileOperations
 
+        // An undefined directory property for use when providers are disabled
+        def nullDirectory = project.objects.directoryProperty()
+
         // Create a task that will be used to merge the task-specific schema locations to the directory (or directories)
         // originally specified.  This allows us to fan out the generated output and keep good cacheability for the
         // compile/kapt tasks but still join everything later in the location the user expects.
@@ -102,7 +105,7 @@ class RoomSchemaLocationWorkaround implements Workaround {
 
                 // Add a command line argument provider to the task-specific list of providers
                 task.options.compilerArgumentProviders.add(
-                    new JavaCompilerRoomSchemaLocationArgumentProvider(roomExtension.schemaLocationDir, taskSpecificSchemaDir, { javaCompileSchemaGenerationEnabled } as Supplier)
+                    new JavaCompilerRoomSchemaLocationArgumentProvider(roomExtension.schemaLocationDir, taskSpecificSchemaDir, nullDirectory, { javaCompileSchemaGenerationEnabled } as Supplier)
                 )
 
                 // Register the generated schemas to be merged back to the original specified schema directory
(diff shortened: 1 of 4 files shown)


src/test/groovy/org/gradle/android/RoomSchemaLocationWorkaroundTest.groovy
```diff
@@ -238,6 +239,77 @@ class RoomSchemaLocationWorkaroundTest extends AbstractTest {
         kotlinVersion << [ "1.3.61", "1.3.50" ].collect { VersionNumber.parse(it) }
     }
 
+    @Unroll
+    def "schemas are correctly generated when only one variant is built incrementally (Android #androidVersion)"() {
+        SimpleAndroidApp.builder(temporaryFolder.root, cacheDir)
+            .withAndroidVersion(androidVersion)
+            .withKotlinVersion(VersionNumber.parse("1.5.0"))
```
runningcode, line 246: we should be consistent with the other tests here and test all kotlin versions or use a variable here or use something like `Versions.latestForKotlin()`

src/test/groovy/org/gradle/android/RoomSchemaLocationWorkaroundTest.groovy
```diff
@@ -382,7 +450,40 @@ class RoomSchemaLocationWorkaroundTest extends AbstractTest {
...
+        assertRoomSchemaContainsColumn("app", "build/roomSchemas/kapt${variant.capitalize()}Kotlin", columnName)
+        assertRoomSchemaContainsColumn("library", "build/roomSchemas/kapt${variant.capitalize()}Kotlin", columnName)
+    }
+
+    void assertMergedRoomSchemaContainsColumn(String columnName) {
+        assertRoomSchemaContainsColumn("app", 'schemas', columnName)
+        assertRoomSchemaContainsColumn("library", 'schemas', columnName)
+    }
+
+    void assertRoomSchemaContainsColumn(String project, String baseDirPath, String columnName) {
```
runningcode, line 476: oh nice. we even check the column exists.

src/test/groovy/org/gradle/android/RoomSchemaLocationWorkaroundTest.groovy
```diff
@@ -382,7 +450,40 @@ class RoomSchemaLocationWorkaroundTest extends AbstractTest {
...
+        assert file("${roomSchemaDirPath(project, 'schemas')}/2.json").text.contains("\"columnName\": \"${columnName}\",")
```
runningcode, line 477: should we also check the the columnName doesn't exist int the `1.json` to make sure files didn't get crossed?

src/main/groovy/org/gradle/android/workarounds/RoomSchemaLocationWorkaround.groovy
```diff
@@ -148,27 +158,24 @@ class RoomSchemaLocationWorkaround implements Workaround {
...
+                }
             }
 
             project.tasks.withType(kaptWithoutKotlincTaskClass).configureEach(configureKaptTask)
             project.tasks.withType(kaptWithKotlincTaskClass).configureEach(configureKaptTask)
 
-            // Since we've added a new kapt-specific provider to the variant, go through the
-            // JavaCompile tasks and remove this provider from its task-specific list so that
-            // it only has its JavaCompile-specific provider.  This is not great, but there
+            // Since we've added a new kapt-specific provider to the variant, disable the provider
```
runningcode, line 174: I honestly am not sure why the javacompile task runs here. I think it might be a mistake if the kapt task is the one generating the schema. Do you think we should report this to Jetbrains?

runningcode (approved):
Looks good aside from the comments! Cheers!

src/test/groovy/org/gradle/android/RoomSchemaLocationWorkaroundTest.groovy
ghale: Good point.  I changed it to just use latest version - I don't think this test really has any particular sensitivity to kotlin, so I think it's fine to just run it against the one version.

src/test/groovy/org/gradle/android/RoomSchemaLocationWorkaroundTest.groovy
ghale: I've added a content check to the overall "schemas are merged properly" check to ensure that 1.json gets copied over unchanged.

src/main/groovy/org/gradle/android/workarounds/RoomSchemaLocationWorkaround.groovy
runningcode: Hmm, I thought they were both compiled by the kotlin compiler since if a java file references a kotlin file, the java compiler would not know what to do with it.
For annotation processing this is achieved by the kotlin compiler creating java stubs for kotlin files but this is not done if there are no annotation processors.

PR #191, opened 2021-08-24
Relax version check to allow new patch versions of supported Android versions

ghale:


runningcode:
This looks good. So just to confirm: if I try to use this with AGP 7.1 it will throw the exception but if I try to use it with AGP 7.0.2 which isn't in the list of supported versions, it will display a warning. With 7.0.1 it will work with no warning.

runningcode (approved):

ghale:
Yeah, that's the idea - if it exactly matches a supported version, no warning; if it's just a different patch version of a supported version, you get a warning; if it's a different minor version from any supported version, you get an error.  The assumption is that nothing will change drastically between patch versions.  I suspect this might be painful for alpha versions where that assumption may not hold true, but I guess if you're riding the alpha wave, you're accepting a certain degree of uncertainty.

PR #203, opened 2021-10-05
Fix non-relocatable build caching for test task

rnett:
Does what the title says and makes the `test` task fully cachable.  It relativizes the local repository using the project dir, and explicitly uses the project dir as the working dir.


runningcode (approved):
Nice find and fix!

PR #212, opened 2021-11-15
Disable cache for ZipMergingTask

jprinet:


src/main/groovy/org/gradle/android/workarounds/ZipMergingTaskWorkaround.groovy
```diff
@@ -0,0 +1,27 @@
...
+
+import com.android.build.gradle.tasks.ZipMergingTask
+import org.gradle.android.AndroidIssue
+import org.gradle.api.Project
+
+/**
+ * Disables caching of ZipMergingTask which is mostly disk bound and
+ * unlikely to provide positive performance benefits.
+ */
+@AndroidIssue(introducedIn = "3.5.4", fixedIn = [], link="https://issuetracker.google.com/issues/200002454")
```
runningcode, line 11:
```suggestion
@AndroidIssue(introducedIn = "3.5.0", fixedIn = [], link="https://issuetracker.google.com/issues/200002454")
```
Just in case we have some customers using the cache fix plugin with 3.5.0!

src/main/groovy/org/gradle/android/workarounds/ZipMergingTaskWorkaround.groovy
```diff
@@ -0,0 +1,27 @@
...
+class ZipMergingTaskWorkaround implements Workaround {
+
+    private static final String CACHING_ENABLED_PROPERTY = "org.gradle.android.cache-fix.ZipMergingTask.caching.enabled"
+
+    @Override
+    void apply(WorkaroundContext context) {
+        context.project.tasks.withType(ZipMergingTask).configureEach {
+            outputs.doNotCacheIf("Caching ZipMergingTask is unlikely to provide positive performance results.", {true })
```
runningcode, line 19:
```suggestion
            outputs.doNotCacheIf("Caching ZipMergingTask is unlikely to provide positive performance results.", { true })
```

src/test/groovy/org/gradle/android/ZipMergingTaskWorkaroundTest.groovy
```diff
@@ -0,0 +1,51 @@
+package org.gradle.android
+
+import org.gradle.testkit.runner.BuildResult
+import org.gradle.testkit.runner.TaskOutcome
+import spock.lang.Unroll
+
+@MultiVersionTest
+class ZipMergingTaskWorkaroundTest extends AbstractTest {
```
runningcode, line 8: Thanks for adding the test. I wonder if we can add additional tasks to execute to the existing `CrossVersionOutcomeAndRelocationTest`. @ghale what do you think?

runningcode (commented):
Looks good, just a comment about the additional test.

src/test/groovy/org/gradle/android/ZipMergingTaskWorkaroundTest.groovy
jprinet: Yep didn't want to mess up the existing `CrossVersionOutcomeAndRelocationTest` but if we have more tasks not in the assemble task graph, it would probably be better to incorporate them there.

ghale: Yeah, this feels like something that should go into `CrossVersionOutcomeAndRelocationTest`.  We have dedicated tasks for workarounds that have some sort of special behavior, but since this one just enables/disables caching, I don't think we need a separate test.  Am I understanding correctly that this task is not captured by `assemble`?  If so, then I'd say let's either add the `createFullJar` tasks to the command line or change the `SimpleAndroidApp` build scripts to add these as dependencies to `assemble`.

jprinet: You're absolutely right @ghale, I'll add the task as a dependency to `assemble` to make this smooth.
great suggestion 👍

src/test/groovy/org/gradle/android/CrossVersionOutcomeAndRelocationTest.groovy
```diff
@@ -542,6 +544,8 @@ class CrossVersionOutcomeAndRelocationTest extends AbstractTest {
         // Renamed from ToJar to ToDir
```
jprinet, line 544: I had to add some specific expectations as you can see below, the comment is probably obsolete then

src/test/groovy/org/gradle/android/SimpleAndroidApp.groovy
```diff
@@ -113,7 +113,10 @@ class SimpleAndroidApp {
                 }
             """.stripIndent()
 
-        file("${library}/build.gradle") << subprojectConfiguration("com.android.library") << activityDependency()
+        file("${library}/build.gradle") << subprojectConfiguration("com.android.library") << activityDependency() <<
+            """
+                tasks['assemble'].dependsOn 'createFullJarDebug', 'createFullJarRelease'
```
runningcode, line 118: instead of adding a task dependency, can we add `createFullJarDebug` and createFullJarRelease` on the command line?
does adding `createFullJar` to the command line arguments achieve the same thing?

src/test/groovy/org/gradle/android/CrossVersionOutcomeAndRelocationTest.groovy
```diff
@@ -48,13 +48,13 @@ class CrossVersionOutcomeAndRelocationTest extends AbstractTest {
 
         withGradleVersion(gradleVersion.version)
             .withProjectDir(originalDir)
-            .withArguments("assemble", "--build-cache", "--stacktrace")
+            .withArguments("assemble", "createFullJarDebug", "createFullJarDebug", "--build-cache", "--stacktrace")
```
ghale, line 51: I think this should be `"createFullJarDebug", "createFullJarRelease"`

src/test/groovy/org/gradle/android/CrossVersionOutcomeAndRelocationTest.groovy
```diff
@@ -48,13 +48,13 @@ class CrossVersionOutcomeAndRelocationTest extends AbstractTest {
...
             .build()
 
         when:
         def result = withGradleVersion(gradleVersion.version)
             .withProjectDir(relocatedDir)
-            .withArguments("assemble", "--build-cache", "--stacktrace")
+            .withArguments("assemble", "createFullJarDebug", "createFullJarDebug", "--build-cache", "--stacktrace")
```
ghale, line 57: Same here

ghale (changes requested):

src/test/groovy/org/gradle/android/CrossVersionOutcomeAndRelocationTest.groovy
ghale: Yeah, let's remove this.  It's not really important and not technically correct.

src/test/groovy/org/gradle/android/CrossVersionOutcomeAndRelocationTest.groovy
jprinet: The evil copy/paste striked again 😅 thanks for spotting!

src/test/groovy/org/gradle/android/SimpleAndroidApp.groovy
jprinet: done, it might make more sense indeed

ghale (approved):

PR #213, opened 2021-11-29
Remove issues that have now been resolved

bigdaz:
* DexFileDependenciesTask: was made cacheable in `AGP 7.0.0-alpha05`
* CompileLibraryResourcesTask: experimental flags to fix absolute resource paths were enabled by default from `AGP 7.2-beta04`

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,12 +77,10 @@ static class AndroidJavaCompile_BootClasspath_Workaround implements Workaround {
 
 ### Unresolved Issues
 
-The following caching issues are fixed by the cache fix plugin but unresolved in any current or upcoming preview release of the Android Gradle Plugin as of 21.08.2020.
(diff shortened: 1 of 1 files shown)


runningcode (approved):
Thanks for updating this!

We'll need to do a follow up to the `CompileLibraryResourcesTaskWorkaround` so that it doens't print this in AGP 7.1: https://github.com/gradle/android-cache-fix-gradle-plugin/blob/master/src/main/groovy/org/gradle/android/workarounds/CompileLibraryResourcesWorkaround_7_0.groovy

barnhill:
CompileLibraryResourcesTask is not relocatable: https://issuetracker.google.com/issues/155218379

Has apparently been fixed according to Google as mentioned in the issue tracker linked above.  Starting with 7.2.0-beta01

PR #242, opened 2022-02-14
Remove init script.

runningcode:
Remove init script in favor of directly applying GE plugin.


runningcode:
@ghale do we still need the init script for debugging tests?

etiennestuder:
If we still need it for debugging, let's store it somewhere internally but not in this repo. It's confusing to have such an init script that instruments the build for GE.

ghale:
Echo what Etienne said.  Let's just get rid of it.  We can always handle test debugging separately.

ghale (approved):
LGTM

PR #248, opened 2022-02-28
Extract build logic

alextu:
This PR refactors a bit the build:

1. Workflow file

    - Remove all comments
    - Rename/add step names
    - Upgrade Github Actions versions
    - file will be renamed just before merging (to be able to still run tests during the review)

2. Gradle Build logic

    There's a version mapping between the supported Android versions and Gradle versions. This is currently used to: 
    - run one test task per specific Android version, on Github Actions all tasks being run with a matrix job
    - generate a versions.json file embedded and used in the plugin itself
   
    To better surface those usages, this logic is extracted in an included build with:
     - Domain object to express the supported Android versions -> Gradle versions
     - Github action specific task to generate matrix in stdout (instead of json file)


runningcode:
I'm not sure moving the version logic to `build-logic` makes this simpler or easier to understand.

build-logic/build.gradle
```diff
@@ -0,0 +1,22 @@
...
+
+dependencies {
+    implementation 'com.fasterxml.jackson.core:jackson-core:2.13.1'
+    implementation 'com.fasterxml.jackson.core:jackson-databind:2.13.1'
+}
+
+gradlePlugin {
+    plugins {
+        androidVersionsPlugin {
+            id = 'org.gradle.android.plugin'
```
ghale, line 18: This doesn't seem like a very descriptive plugin id.  I'd make this `org.gradle.android-versions` or something like that.

build-logic/src/main/java/org/gradle/android/plugin/AndroidVersionsPlugin.java
```diff
@@ -0,0 +1,62 @@
...
+        var androidVersionsContainer = objects.domainObjectContainer(AndroidVersionDomainObject.class,
+            name -> objects.newInstance(AndroidVersionDomainObject.class, name));
+        project.getExtensions().add("androidVersions", androidVersionsContainer);
+        var generateTask = project.getTasks().register("generateVersions", GenerateVersionsTask.class);
+        generateTask.configure(t -> {
+            t.getSupportedVersions().set(androidVersionsContainer);
+            t.getVersion().set((String) project.getVersion());
+        });
+
+        var sourceSets = project.getExtensions().getByType(JavaPluginExtension.class).getSourceSets();
```
ghale, line 27: This plugin expects `JavaPluginExtension` to exist in order to work properly.  It should either apply the java plugin, or it should respond to the java plugin being applied (i.e. `pluginManager.withPlugin()`).  Otherwise, you have an ordering problem (i.e. java plugin _must_ be applied before this plugin in the applying project).

build-logic/src/main/java/org/gradle/android/plugin/AndroidVersionsPlugin.java
```diff
@@ -0,0 +1,62 @@
...
+            project.getTasks().register("generateGithubActionsMatrix", GithubActionsMatrixTask.class).configure(t -> t.getTasks().set(tasks));
+        });
+    }
+
+    private TaskProvider<Test> registerAndroidTestTask(Project project, AndroidVersionDomainObject androidVersion) {
+        var versionSpecificTest = project.getTasks().register(androidTestTaskName(androidVersion.getName()), Test.class);
+        versionSpecificTest.configure(t -> {
+            t.setDescription("Runs the multi-version tests for AGP " + androidVersion.getName());
+            t.setGroup("verification");
+            t.systemProperty("org.gradle.android.testVersion", androidVersion.getName());
```
ghale, line 48: It feels weird that now we have this reusable piece of logic (that generates android test tasks) that also depends on the project containing a `SpockConfig.groovy` that sets up the categories.  Perhaps we should pull `SpockConfig.groovy` and `MultiVersionTest` into the plugin, compile them as a separate sourceset and add them to the test classpath?  Or maybe solve this problem some other way completely (for instance having a separate test source set for multiversion tests)?

ghale (changes requested):
Pragmatically, I think this is fine as is.  However, I've commented on some things that feel very awkward about this.  I think we should consider what we are gaining by making this change.  Typically, the value of pulling build logic out like this is 1) being able to test it like any other software artifact and 2) being able to eventually reuse it elsewhere (either in a multiproject build or as a reusable plugin across distinctly different builds).  I see neither of those at the moment.  If we want to add tests, then there's probably some value (we could do this as a follow up task, doesn't have to be in this PR).  I don't see any immediate opportunities to reuse this anywhere, though.

Let's first be clear about what we want to accomplish with this change before we decide to move forward with this.

.github/workflows/ci-check.yml
```diff
@@ -1,92 +1,65 @@
...
-      fail-fast: false
-      matrix:
-        androidTestTask: ${{ fromJson(needs.generate_versions.outputs.matrix) }}
-
-    # Steps represent a sequence of tasks that will be executed as part of the job
-    steps:
-    # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
-    -   uses: actions/checkout@v1
-
-    # Patch issue in platform-tools 31.0.3 where platform-tools/api/api-versions.xml is missing (see https://issuetracker.google.com/issues/195445762)
```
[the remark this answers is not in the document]
etiennestuder: Agreed. Remove `all` comments was a bit too strong of a goal. Remove `redundant` would be more appropriate.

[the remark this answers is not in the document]
etiennestuder: @runningcode did the 'remove most comments' on his other PR.

build-logic/build.gradle
etiennestuder: Agreed.

build-logic/src/main/java/org/gradle/android/plugin/AndroidVersionsPlugin.java
etiennestuder: Agreed.

build-logic/src/main/java/org/gradle/android/plugin/AndroidVersionsPlugin.java
etiennestuder: Let's pick up the conversation on this again once we have a new PR available that soley focusses on the multi-version aspect.

etiennestuder:
I'd like to close this PR since it went a bit further than anticipated. I suggest that we focus on the original intention of improving/simplifyimng the build/yml in terms of not having to run tasks to generate a json file with test task names. This can be done in a simpler way, unless I'm totally overlooking something.

To me, it seems that the source of truth are the versions to support and test. If this is the case, I would move the definition of these versions to a json file. From there, it can be read by whoever needs to read it (the build, the production code, GH action). As a consequence, there would no longer be the need for a setup task, the versions are easily identifiable by looking at the json file, and the sequence of events would be simpler and easier to understand, both when looking at the build and when looking at the GH action yml file.

//cc @runningcode @ghale @alextu

PR #251, opened 2022-03-02
Clean up Github Action workflow

runningcode:
Cleans up the Github action workflow comments

* Updates to checkout v2 and java v2
* Adds step names
* Use java 11 for tests (java 8 is no longer required)

### DIFF
--- a/.github/workflows/ci-check.yml
+++ b/.github/workflows/ci-check.yml
@@ -1,9 +1,6 @@
-# This is a basic workflow to help you get started with Actions
+## TODO RENAME THIS FILE TO build-verification.yml BEFORE MERGING (KEPT SAME FOR EASIER REVIEW)
+name: Verify Build
 
-name: CI
-
-# Controls when the action will run. Triggers the workflow on push or pull request
-# events but only for the master branch
 on:
   push:
     branches: [ master, citest ]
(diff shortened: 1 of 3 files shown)


runningcode:
No whitespace for easier review: https://github.com/gradle/android-cache-fix-gradle-plugin/pull/251/files?diff=split&w=1

This takes most of the cleanup workflow cleanup from this PR https://github.com/gradle/android-cache-fix-gradle-plugin/pull/248

I will rename the workflow before merging. The name is currently kept the same for easier review.

ghale (approved):
LGTM

.github/workflows/ci-check.yml
```diff
@@ -1,92 +1,61 @@
...
-    -   uses: actions/setup-java@v1
+      - uses: actions/setup-java@v2
         with:
-            java-version: 8
-
-    # Gradle build
-    -   uses: gradle/gradle-build-action@v2
-        env:
-            GRADLE_ENTERPRISE_ACCESS_KEY: ${{ secrets.GE_ACCESS_TOKEN }}
+          java-version: 11
```
etiennestuder, line 58: Is the move from Java 8 to Java 11 intentional?

etiennestuder (approved):
Added one question.

.github/workflows/ci-check.yml
ghale: I assumed that it is.  We used to build with JDK8 because older versions of AGP required it.  So, we would build with JDK8, test those versions with JDK8, and test against newer versions of AGP using JDK11.  With the work to drop support for older versions of AGP we did recently, we now only support versions that require JDK11 and up.  As far as I can tell, there's no reason to build with JDK8 anymore.

runningcode: Good question. Yes, this is intentional. AGP 7+ only supports Java 11+.
The build also has the toolchain version set to 11 https://github.com/gradle/android-cache-fix-gradle-plugin/blob/main/build.gradle#L61

PR #252, opened 2022-03-04
Add support for AGP 7.2.0-beta03

runningcode:


etiennestuder (changes requested):
Added one question on why the access key is not defined on the job level.

runningcode:
I adjusted the `env` jars to be on the workflow level.

PR #255, opened 2022-03-08
Move versions out to an external json file

ghale:
- Move supported version information to an external json file
- Generate test tasks from json file at build config time
- Use json file to generate test matrix in the GH action


src/test/groovy/org/gradle/android/SimpleAndroidApp.groovy
```diff
@@ -58,7 +64,7 @@ class SimpleAndroidApp {
                     }
                     dependencies {
                         classpath ('com.android.tools.build:gradle:$androidVersion') { force = true }
-                        classpath "org.gradle.android:android-cache-fix-gradle-plugin:${Versions.PLUGIN_VERSION}"
```
runningcode, line 61: should we remove `Versions.PLUGIN_VERSION` or make it point to the system property?

runningcode (commented):
just a single comment otherwise looks good. this very much simplifies the build.

src/test/groovy/org/gradle/android/SimpleAndroidApp.groovy
ghale: Well Versions is part of the production code.  `Versions.PLUGIN_VERSION` is only used by the test fixtures.  And it requires configuration cache gymnastics in order to read the system property safely.  Rather than introduce that, it seemed cleaner to just remove it from the production code and handle it in the test fixture as a test runtime specific system property.

runningcode (approved):
Thanks!

PR #256, opened 2022-03-09
Use CompileStatic

runningcode:
Note: this does not add CompileStatic for the RoomSchemaLocationWorkaround


ghale (approved):
This all LGTM.  There's a minor conflict with the Groovy upgrade I merged earlier today, but it should be trivial to resolve.

etiennestuder:
@runningcode Would you mind adding a comment what the motivation for this change was? Just so everyone reading the PR understands why we are doing this.

PR #257, opened 2022-03-09
Simplify expected outcomes

ghale:
This simplifies the expected outcomes in the `CrossVersionOutcomeAndRelocation` test.  Rather than have a complicated set of conditionals and version constraints, each version has a json file with the expected outcomes.  This has the advantage that the outcomes are easier to reason about and compare between versions.  It does mean these files have to be maintained and a new file created for new versions, but I think this will be low effort as you can simply copy or rename a file from the previous version and then update the tasks which are different.


runningcode (approved):
This looks great! I like how each version is in a separate outcome.json file.

PR #263, opened 2022-03-14
Add support for AGP 7.3

### LINKED ISSUE #262: Test min and max Gradle versions in compatibility matrix
We should ensure that the minimum supported Gradle version and the max supported Gradle version are properly tested.

See here
https://github.com/gradle/android-cache-fix-gradle-plugin/blob/main/src/main/resources/versions.json#L4

runningcode:
Also test against lowest and highest supported Gradle versions.

The `app:mergeXXXResources` tasks are now cacheable compared to AGP 7.2.

Fixes #262


src/test/groovy/org/gradle/android/ConfigurationCachingTest.groovy
```diff
@@ -22,7 +22,10 @@ class ConfigurationCachingTest extends AbstractTest {
                 .build()
 
         then:
-        !result.output.contains("problems were found storing the configuration cache")
+        // In Gradle 7.4 the build always uses the following words.
+        if (!result.output.contains("0 problems were found storing the configuration cache")) {
+            !result.output.contains("problems were found storing the configuration cache")
+        }
```
ghale, line 28: I find this conditional hard to reason about.  It might be better to use a method so that the intention is really clear - something like:
```
static final String CC_PROBLEMS_FOUND = "problems were found storing the configuration cache"

...

then:
assertNoConfigCacheProblemsFound(result)

...

void assertNoConfigCacheProblemsFound(BuildResult result) {
    assert !result.output.contains(CC_PROBLEMS_FOUND) || result.output.contains("0 ${CC_PROBLEMS_FOUND}")
}
```

ghale (approved):
One minor comment on the test, but otherwise looks good.

src/test/groovy/org/gradle/android/ConfigurationCachingTest.groovy
runningcode: This is a good point. Thanks for the clear code snippet!

PR #264, opened 2022-03-16
Fix missing signature files when publishing to the plugin portal

alextu:


etiennestuder (approved):
LGTM. Some of the GH action builds are failing, not sure if it is related to your change or not.

alextu:
> LGTM. Some of the GH action builds are failing, not sure if it is related to your change or not.

Unfortunately it looks related, will investigate

runningcode:
A very quick glance seems to indicate that the publication to the mavenLocal repository which is used by the test isn't working right.

etiennestuder:
I will bail out of the review here. @ghale and @runningcode are better suited to verify this.

alextu:
> A very quick glance seems to indicate that the publication to the mavenLocal repository which is used by the test isn't working right.

Yes, I've just made a change that should fix it. Will wait for the "Verify Build" to confirm.

build.gradle
```diff
@@ -80,6 +82,15 @@ pluginBundle {
 // in the TestKit plugin classpath.
 def localRepo = file("$buildDir/local-repo")
 publishing {
+    publications {
+        maven(MavenPublication) {
```
runningcode, line 86: Can you add a comment here if we can remove this section in the future when the plugin portal automatically supports publishing the signature files? That way we aren't wondering if this should be removed.

runningcode (approved):

runningcode:
Looks good, thanks!

build.gradle
alextu: done

PR #273, opened 2022-04-06
Fix issue with spurious CompileLibraryResourcesWorkaround warnings

### LINKED ISSUE #234: CompileLibraryResourcesTask warning even when experimental properties are set
I use version 2.4.6 of this plugin, and enable the experimental properties that AGP added in `gradle.properties` for both my main project and an included build:

```
android.experimental.cacheCompileLibResources=true
android.experimental.enableSourceSetPathsMap=true
```

It seems like when the included build is built completely from scratch (not incremental or in the build cache), this warning gets printed anyways.

```
> Task :my-included-build:generatePrecompiledScriptPluginAccessors
WARNING: Android plugin 7.2.0-beta01 has experimental support for using relative path sensitivity with CompileLibraryResourcesTask inputs which will provide more build cache hits and improve build speed.  Set 'android.experimental.cacheCompileLibResources=true' and 'android.experimental.enableSourceSetPathsMap=true' in gradle.properties to enable this support.
Android plugin 7.2.0-beta01 has experimental support for using relative path sensitivity with CompileLibraryResourcesTask inputs which will provide more build cache hits and improve build speed.  Set 'android.experimental.cacheCompileLibResources=true' and 'android.experimental.enableSourceSetPathsMap=true' in gradle.properties to enable this support.
```

ghale:
When a Kotlin script plugin is compiled, Gradle runs a nested build with the applied plugins in order to inspect/generate the type-safe accessors for the script.  When a Kotlin script plugin applies the cache-fix plugin, the plugin is applied to the nested build without inheriting the Gradle properties from the main build.  This causes the warnings from workarounds like `CompileLibraryResourcesWorkaround` to get printed to the console, even though the properties that it references are actually applied in the main project.

This fixes the issue by inhibiting warnings when we detect that we're in the nested build for generating type-safe accessors.  We detect this by just looking at the root project name, so there is of course the possibility of another project that just happens to be named `gradle-kotlin-dsl-accessors` potentially having their warnings inhibited, but I think this is extremely low probability.

Fixes #234

### DIFF
--- a/src/main/groovy/org/gradle/android/AndroidCacheFixPlugin.groovy
+++ b/src/main/groovy/org/gradle/android/AndroidCacheFixPlugin.groovy
@@ -61,7 +61,7 @@ class AndroidCacheFixPlugin implements Plugin<Project> {
 
         if (!isSupportedAndroidVersion(project)) {
             if (isMaybeSupportedAndroidVersion(project)) {
-                Warnings.MAYBE_SUPPORTED_ANDROID_VERSION.warnOnce(project.logger)
+                Warnings.MAYBE_SUPPORTED_ANDROID_VERSION.warnOnce(project)
             } else {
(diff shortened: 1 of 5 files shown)


runningcode (approved):
Very interesting finding and fix! Thanks. Good to keep in mind the caveat that the warnings won't be printed when applied from a kotlin script plugin.

PR #282, opened 2022-04-13
Remove CompileLibraryResourcesWorkaround in AGP 7.2.0-beta01+

### LINKED ISSUE #281: CompileLibraryResourcesWorkaround should not be needed anymore starting with AGP 7.2-beta01
According to: https://issuetracker.google.com/issues/155218379#comment6

runningcode:
Fixes #281

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@
 Some Android plugin versions have issues with Gradle's build cache feature. When applied to an Android project this plugin applies workarounds for these issues based on the Android plugin and Gradle versions. For other versions, please see [older versions.](#older-android-gradle-plugin-versions)
 
 * Supported Gradle versions: 7.0+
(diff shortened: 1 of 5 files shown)


vanniktech (approved):
Nice!

These might be fixed too:

https://github.com/gradle/android-cache-fix-gradle-plugin/blob/b8001fb0d1e25a6fa17483d3bb10f3ddb56fcdf2/src/main/groovy/org/gradle/android/workarounds/ZipMergingTaskWorkaround.groovy#L13
https://github.com/gradle/android-cache-fix-gradle-plugin/blob/b8001fb0d1e25a6fa17483d3bb10f3ddb56fcdf2/src/main/groovy/org/gradle/android/workarounds/MergeNativeLibsWorkaround.groovy#L9

Although I haven't quite found out for which version. From the comment it seems like `7.2.0-alpha06` but there's no reference to gerrit or anything else.

ghale (approved):
LGTM

PR #283, opened 2022-04-14
Enable configuration caching

clayburn:


kozaxinan (approved):

build.gradle
```diff
@@ -186,6 +186,15 @@ tasks.register('createReleaseTag', CreateGitTag) {
...
 tasks.named("githubRelease") {
     dependsOn("createReleaseTag")
+    notCompatibleWithConfigurationCache("$name task does not support configuration caching")
+}
+
+tasks.named('publishPlugins') {
+    notCompatibleWithConfigurationCache("$name task does not support configuration caching")
+}
+
+tasks.withType(Sign) {
```
eskatos, line 196: This would force the configuration of all `Sign` tasks in all builds. The following would keep the configuration lazy:
```suggestion
tasks.withType(Sign).configureEach {
```

clayburn: Thanks, I've made the change.

build.gradle
```diff
@@ -186,6 +186,15 @@ tasks.register('createReleaseTag', CreateGitTag) {
 
 tasks.named("githubRelease") {
     dependsOn("createReleaseTag")
+    notCompatibleWithConfigurationCache("$name task does not support configuration caching")
```
etiennestuder, line 189: Same comment as on the CCUD plugin.

etiennestuder (changes requested):

build.gradle
```diff
@@ -186,6 +186,15 @@ tasks.register('createReleaseTag', CreateGitTag) {
 
 tasks.named("githubRelease") {
```
runningcode, line 187:
```suggestion
tasks.named("githubRelease").configure {
```
and for the others too to make it use the task configuration avoidance api

etiennestuder: @runningcode on this one, I was not sure if it is needed to get lazy task instantiation. It's defintely needed?

etiennestuder: If so, we would also have to do it for `publishPlugins` (and on the other 3 repos where we have PRs open).

runningcode: I think `tasks.named` is ok to use according to the [task configuration avoidance guide](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html).

That being said, it is currently [being eagerly configured](https://ge.solutions-team.gradle.com/s/2ijgwbnylh2cy/performance/configuration?details=code-unit-net.breadmoirai.github-release!:&openScriptsAndPlugins=WyJjb2RlLXVuaXQtbmV0LmJyZWFkbW9pcmFpLmdpdGh1Yi1yZWxlYXNlIl0) which is why I brought it up. I think that is an issue in the plugin itself and not this buildscript.

Conclusion: we can leave out `.configure` here.

etiennestuder: Got it. I guess it does not hurt to have it here.

PR #284, opened 2022-04-15
Fix group when publishing on staging portal

alextu:
The var was set to null instead of the expected 'org.gradle.android', causing the group to eventually be set with the gradle.plugin prefix.


jprinet (approved):
LGTM

PR #295, opened 2022-04-30
Add a workaround for the JDK Image cache miss issue 

ghale:
This adds a workaround for cache misses caused by the custom Java runtime used when source compatibility is set to a version higher than Java 9.  The workaround does the following:

- Replaces the command line provider with one that uses the extracted runtime image as its input, which allows further runtime normalization if necessary.
- Normalizes out the version number in `module-info.class`
- Adds normalization of `Implementation-Version`, `Implementation-Vendor` and `Created-By` in manifest properties.

The net result of this is that as long as the classes in the custom runtime are the same, we should get a cache hit, even if there are minor differences in the JDKs used to create the runtime.  It's also possible to do further normalization if there are minor differences not accounted for by the workaround.


runningcode:
If this is ready, let's remove the `WIP` in the title :)

PR #314, opened 2022-07-04
Fix a failure when ZULU_ env vars are not set

alextu:
The new plugin-publish plugin somehow configures the test task on publishing.


runningcode (approved):
LGTM

PR #325, opened 2022-07-26
Allow RoomSchemaLocationWorkaround to be disabled

cdsap:
Implements the ability to suppress RoomSchemaLocationWorkaround workarounds


ghale (approved):
LGTM. Don't forget to add the new feature to `release/changes.md`.  You can look at the history over the past few releases to see the standard format we use.

PR #330, opened 2022-08-09
Replace jcenter with mavenCentral

jeremydyoung:
We discovered that the tests are still building projects using jcenter, so this is a simple update to swap that out.  Hopefully all works correctly.


cdsap (approved):

ghale:
LGTM, thanks @jeremydyoung!

PR #347, opened 2022-09-08
Suggest updating plugin when current AGP version is not supported

### LINKED ISSUE #345: Suggest updating plugin when unsupported AGP is detected
When updating to a newer version of AGP it is common that the version of android-cache-fix-gradle-plugin also needs to be updated.

Right now the error message there only suggests using an older version of AGP or suppressing the warning.

Since Gradle seems to be pretty well on top of releasing new versions of the plugin when a new stable release of AGP drops it would be nice to let the first suggestion be to check for an update of this plugin.

jprinet:
Fixes #345


runningcode (approved):
I have a feeling some tests might fail but otherwise this looks great. Thanks for doing this quickly :)

PR #356, opened 2022-10-14
Support new variant API on AGP 7.4.0 and up

### LINKED ISSUE #353: Android Gradle Plugin 7.4.0-beta02 fails with android-cache-fix-gradle-plugin
Environment :
  Gradle : 7.5.1
  AGP : 7.4.0-beta02
  android-cache-fix-gradle-plugin : 2.5.6
  Java : OpenJDK 11

How to reproduce failure
```
git clone https://github.com/ganadist/VersionCodeDemo -b agp74_cache_fix
cd VersionCodeDemo
./gradlew :lib:glide:assemble

> Configure project :lib:glide
WARNING: Android plugin 7.4.0-beta02 has not been tested with this version of the Android cache fix plugin, although it may work.  We test against only the latest patch release versions of Android Gradle plugin: 7.0.4, 7.1.3, 7.2.1, 7.3.0-beta05, 7.4.0-alpha08.  If 7.4.0-beta02 is newly released, we may not have had a chance to release a version tested against it yet.  Proceed with caution.  You can suppress this warning with with -Dorg.gradle.android.cache-fix.ignoreVersionCheck=true.
Values of variant API AnnotationProcessorOptions.arguments are queried and may return non final values, this is unsupported

FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring project ':lib:glide'.
> com.android.build.gradle.internal.crash.ExternalApiUsageException: com.android.build.gradle.internal.dsl.AgpDslLockedException: It is too late to modify AnnotationProcessor::argumentProviders
  It has already been read to configure this project.
  Consider either moving this call to be during evaluation,
  or using the variant API.


Caused by: com.android.build.gradle.internal.dsl.AgpDslLockedException: It is too late to modify AnnotationProcessor::argumentProviders	
It has already been read to configure this project.	
Consider either moving this call to be during evaluation,	
or using the variant API.	
	at com.android.build.gradle.internal.dsl.decorator.LockableList.add(LockableList.kt:83)	
		•••
	at org.gradle.android.workarounds.RoomSchemaLocationWorkaround$_apply_closure3$_closure14.doCall(RoomSchemaLocationWorkaround.groovy:137)	
		•••
	at com.android.build.gradle.LibraryExtension.addVariant(LibraryExtension.kt:102)	
	at com.android.build.gradle.internal.ApiObjectFactory.create(ApiObjectFactory.java:119)	
	at com.android.build.gradle.internal.plugins.BasePlugin.createAndroidTasks(BasePlugin.kt:706)	
	at com.android.build.gradle.internal.plugins.BasePlugin$createTasks$2$1.call(BasePlugin.kt:572)	
	at com.android.build.gradle.internal.profile.NoOpAnalyticsConfiguratorService.recordBlock(NoOpAnalyticsConfiguratorService.kt:53)	
	at com.android.build.gradle.internal.plugins.BasePlugin$createTasks$2.accept(BasePlugin.kt:567)	
	at com.android.build.gradle.internal.plugins.BasePlugin$createTasks$2.accept(BasePlugin.kt:564)	
	at com.android.build.gradle.internal.crash.CrashReporting$afterEvaluate$1.execute(crash_reporting.kt:37)	
	at com.android.build.gradle.internal.crash.CrashReporting$afterEvaluate$1.execute(crash_reporting.kt:35)
```

ghale:
Use the new variant API on AGP versions 7.4.0 and higher.

Fixes #353

### DIFF
--- a/src/main/groovy/org/gradle/android/workarounds/RoomSchemaLocationWorkaround.groovy
+++ b/src/main/groovy/org/gradle/android/workarounds/RoomSchemaLocationWorkaround.groovy
@@ -91,7 +91,7 @@ class RoomSchemaLocationWorkaround implements Workaround {
                     // Make sure that the annotation processor argument has not been explicitly configured in the Android
                     // configuration (i.e. we only want this configured through the room extension)
                     Map<String, String> arguments = variant.javaCompileOptions.annotationProcessorOptions.arguments
-                    errorIfAnnotationProcessorSet(arguments.keySet())
+                    errorIfRoomSchemaAnnotationArgumentSet(arguments.keySet())
 
                     def variantSpecificSchemaDir = project.objects.directoryProperty()
                     variantSpecificSchemaDir.set(getVariantSpecificSchemaDir(project, "compile${variant.name.capitalize()}JavaWithJavac"))
@@ -111,7 +111,7 @@ class RoomSchemaLocationWorkaround implements Workaround {
                     // Make sure that the annotation processor argument has not been explicitly configured in the Android
                     // configuration (i.e. we only want this configured through the room extension
                     MapProperty<String, String> arguments = variant.javaCompilation.annotationProcessor.arguments
-                    errorIfAnnotationProcessorSet(arguments.keySet().get())
+                    errorIfRoomSchemaAnnotationArgumentSet(arguments.keySet().get())
 
                     def variantSpecificSchemaDir = project.objects.directoryProperty()
                     variantSpecificSchemaDir.set(getVariantSpecificSchemaDir(project, "compile${variant.name.capitalize()}JavaWithJavac"))
@@ -225,7 +225,7 @@ class RoomSchemaLocationWorkaround implements Workaround {
         }
     }
 
-    private static void errorIfAnnotationProcessorSet(Set<String> options) {
+    private static void errorIfRoomSchemaAnnotationArgumentSet(Set<String> options) {
         if (options.contains(ROOM_SCHEMA_LOCATION)) {
             throw new IllegalStateException("""${RoomSchemaLocationWorkaround.class.name} cannot be used with an explicit '${ROOM_SCHEMA_LOCATION}' annotation processor argument.  Please change this to configure the schema location directory via the 'room' project extension:
     room {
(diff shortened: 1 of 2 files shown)


src/main/groovy/org/gradle/android/workarounds/RoomSchemaLocationWorkaround.groovy
```diff
@@ -179,6 +225,24 @@ class RoomSchemaLocationWorkaround implements Workaround {
         }
     }
 
+    private static void errorIfAnnotationProcessorSet(Set<String> options) {
```
runningcode, line 228: maybe rename this to `errorIfRoomSchemaSet`?

I say this because it is possible to use the room annotation processor correctly without the schema migrations and testing.

src/main/groovy/org/gradle/android/workarounds/RoomSchemaLocationWorkaround.groovy
```diff
@@ -188,23 +252,32 @@ class RoomSchemaLocationWorkaround implements Workaround {
         }
     }
 
-    private static void applyToAllAndroidVariants(Project project, Closure<?> configureVariant) {
+    private static void applyToAllAndroidVariants(Project project, ConfigureVariants configureVariants) {
+        if (Versions.CURRENT_ANDROID_VERSION <= VersionNumber.parse("7.4.0-alpha01")) {
```
runningcode, line 256: is this new API only available in 7.4? any downsides to using it 7.3 if it is available there?

runningcode (approved):
Just some small comments otherwise looks good. Thanks for the speedy fix and for keeping us future-proof @ghale !

src/main/groovy/org/gradle/android/workarounds/RoomSchemaLocationWorkaround.groovy
ghale: Only what you were communicating the other day that the new api doesn't work properly on some earlier versions.  Not knowing what versions it works properly with or not, I figured we would just use it starting with the version where we know we have issues with the old api.

src/main/groovy/org/gradle/android/workarounds/RoomSchemaLocationWorkaround.groovy
ghale: renamed to `errorIfRoomSchemaAnnotationArgumentSet`

PR #384, opened 2022-12-08
Creating matrix strategy with Java17

cdsap:


.github/workflows/build-verification.yml
```diff
@@ -70,26 +78,28 @@ jobs:
...
       - name: Set up Java
         uses: actions/setup-java@v3
         with:
-          java-version: 11
+          java-version: ${{ matrix.java.jdk }}
           distribution: 'adopt'
       - name: Set up Gradle
         uses: gradle/gradle-build-action@v2
+      - name: Update libncurses5
+        run: sudo apt update && sudo apt-get install libncurses5
```
cdsap, line 102: required to avoid https://ge.solutions-team.gradle.com/s/3iwp5kve57n54/tests/:testAndroid7_2_2/org.gradle.android.CompileLibraryResourcesWorkaroundTest/does%20not%20warn%20about%20experimental%20flags%20when%20applied%20from%20a%20kotlin%20script%20plugin?top-execution=1

.github/workflows/build-verification.yml
```diff
@@ -60,6 +66,8 @@ jobs:
       fail-fast: false
       matrix:
         versions: ${{ fromJson(needs.generate_versions.outputs.matrix) }}
+        java: [ { jdk: 11, jdkImageIssue: 11.0.12 ,jdkImageIssueAlt:  11.0.13 }, { jdk: 17,  jdkImageIssue:  17.0.3+7 ,jdkImageIssueAlt: 17.0.5+8} ]
```
runningcode, line 69: As long as the JDK image issue is tested it with a different minor version, we can consider it covered. I would remove the second test with a different minor version.

.github/workflows/build-verification.yml
```diff
@@ -23,11 +23,14 @@ jobs:
   sanity_check:
     name: Sanity check
     runs-on: ubuntu-latest
+    strategy:
```
runningcode, line 26: IMO, let's keep the sanity check as a quick check without the matrix strategy.

.github/workflows/build-verification.yml
```diff
@@ -60,6 +66,8 @@ jobs:
       fail-fast: false
       matrix:
         versions: ${{ fromJson(needs.generate_versions.outputs.matrix) }}
+        java: [ { jdk: 11, jdkImageIssue: 11.0.12 ,jdkImageIssueAlt:  11.0.13 }, { jdk: 17,  jdkImageIssue:  17.0.3+7 ,jdkImageIssueAlt: 17.0.5+8},  { jdk: 19,  jdkImageIssue:  19.0.3+7 ,jdkImageIssueAlt: 19.0.5+8} ]
```
runningcode, line 69: I think I failed to see the bigger picture here earlier.
The JDKImage tests are part of the `testAndroidXXX` matrix. This means that the entire matrix is launched with the AGP versions and the differing JDKImage versions. In my previous comment I thought it would be redundant to test the different minor versions of the JDK with each JDKImage issue but since they are all part of the same matrix it would be hard to untangle.

I propose just testing JDK 11 and JDK 19 and removing JDK 17. Sorry to make you undo this: for both JDK 11 and JDK 19 let's also test the different minor versions of the JDKImage issue.

PR #395, opened 2022-12-24
Update Kotlin supported versions with 1.7.21

cdsap:
## Include the latest Kotlin version in supportedKotlinVersions

Discussing the update of JDK17(required by AGP 8.0.0-alpha10) and the KspWorkaround with @runningcode, we noticed we are not testing latest Kotlin in the Android Cache Fix Plugin. Before proceeding with the general update of Jdk17 that affects toolchains, infrastructure and compatibility strategies in the jobs, we need to extend the supported Kotlin versions with 1.7.21. Kotlin and, more specifically, kapt are problematic with older Kotlin versions and newer JDKs. 
This PR adds version 1.7.21 to `supportedKotlinVersions` in `TestVersions`. 

However, it was not straightforward:

### Empty sources in TestSamples
The first issue is the new update returns a different task outcome when the unit of compilation is empty:
```
> Task ':app:compileDebugKotlin' was NO_SOURCE but should have been FROM_CACHE
> Task ':app:compileReleaseKotlin' was NO_SOURCE but should have been FROM_CACHE
> Task ':library:compileDebugKotlin' was NO_SOURCE but should have been FROM_CACHE
> Task ':library:compileReleaseKotlin' was NO_SOURCE but should have been FROM_CACHE
```
To solve this issue, in the SampleAndroid builder, we create a Kotlin class in the app and library only if Kotlin is enabled. This action doesn't require rewriting the task outputs because in older Kotlin versions (wrongly?) the outcome was FROM_CACHE, even when the source was empty.

### BundleLibRes outputs in 7.0.4
The previous fix didn't require changing the outputs of the Kotlin compiler tasks, but we have a side effect in AGP 7.0.4.  The task type `BundleLibraryJavaRes` changes the outcome when one of the predecessor tasks(KotlinCompile) includes sources. We didn't observe this issue in newer AGP versions.

### Kotlin metadata incompatibility in RoomSchemaLocationWorkaroundTest
Side effect of adding Kotlin classes in the app and library modules. This issue surfaces the Kotlin binary incompatibility of dependencies brought by recent AGP versions with older Kotlin versions:
```
Module was compiled with an incompatible version of Kotlin. 
The binary version of its metadata is 1.7.1, expected version is 1.5.1.
```
It's a known issue and we were covered for older Kotlin versions: https://github.com/gradle/android-cache-fix-gradle-plugin/blob/main/src/test/groovy/org/gradle/android/RoomSchemaLocationWorkaroundTest.groovy#L23 
Our problem comes from adding Kotlin sources; we just need to update the Version number used in the assumption because the tests affected never worked for a real project using Kotlin 1.5.x.

 ### Memory issue 
Once the functionality of the test was working, I started triggering builds in CI and started getting builds finished with the error `Error: The operation was canceled.`:
https://github.com/gradle/android-cache-fix-gradle-plugin/actions/runs/3768371716/jobs/6406753264#step:11:11073

https://github.com/gradle/android-cache-fix-gradle-plugin/actions/runs/3768371716/jobs/6406753156#step:11:13860

https://github.com/gradle/android-cache-fix-gradle-plugin/actions/runs/3768378223/jobs/6406763678#step:11:20849

https://github.com/gradle/android-cache-fix-gradle-plugin/actions/runs/3768378223/jobs/6406763870#step:11:16651

After some failed attempts to reproduce the issue locally, I tried to understand the issue where the only change was to include an additional Kotlin Version. 
This change affects the tests either using the latest supportedKotlinVersions or including additional builds execution in tests using all  supportedKotlinVersions. Given the type of interruption in the CI execution and the environment where it is executed (unix/container) the next area of investigation pointed out a memory issue. 

The issue was related to the environment where the build tests are executed. I added  https://github.com/cdsap/InfoKotlinProcess to add custom values with the information of the Kotlin processes. The results showed missing isolation of the Kotlin processes in the Gradle runners. For instance this build https://ge.solutions-team.gradle.com/s/ecm4xcpnhtuas/custom-values#L10 we have:
```
  | Kotlin-Process-2629-uptime | 17.87 minutes |  
  | Kotlin-Process-2954-uptime | 16.98 minutes |  
  | Kotlin-Process-7122-uptime | 10.93 minutes |  
  | Kotlin-Process-7777-uptime | 9.38 minutes |  
```

where in the same environment execution, we have 4 Kotlin processes. After analyzing the capacity of those processes It's not required to rely on the parmeters inherited from the Gradle Daemon. For this reason I added a specific configuration for the Kotlin process in the test samples `-Dkotlin.daemon.jvm.options=-Xmx1500m`

This solution doesn't scale if we don't increase the resources of the build environment and isolate the test execution between different build component versions(Not sharing Kotlin processes).


runningcode:
Wow, very nice finds.

Regarding the empty sources, i think that's a nice fix, are you saying that only happens in KGP 1.7.0+?

Regarding the Kotlin metadata incompatibility, would it simplify the code if we dropped support for KGP 1.5? 

Regarding the memory issue, very nice that your plugin came in handy here!

runningcode:
Ooo, looks we'll need to add KGP `1.8.0` support as well. Let's not make this PR more complicated though, but it makes me favor dropping support for older versions like 1.5.0.

cdsap:
> Regarding the empty sources, i think that's a nice fix, are you saying that only happens in KGP 1.7.0+?
Since Kotlin 1.7 if there are empty sources, the outcome is NO-SOURCE instead of FROM-CACHE.

> Regarding the Kotlin metadata incompatibility, would it simplify the code if we dropped support for KGP 1.5?
Created a different PR #401 

The idea is to merge first #401, then this PR, and finally the update/fix of Kotlin 1.8.0

runningcode (approved):
Nice work and once again thanks for the detailed PR description 👍

PR #405, opened 2023-01-05
AGP 8.0.0-alpha11 and JDK17 CI environment

cdsap:
Update of AGP 8.0.0-alpha11.
It could be an easy update but the new version required to build with JDK17...

## Attempt to update toolchain 

One of the alternatives was to bump the project `toolchain` to Java17.  Build worked locally and test execution as well. At the moment to test the plugin deployed locally, projects using Java11 got this error:
```
Incompatible because this component declares an API of a component compatible 
with Java 17 and the consumer needed a runtime of a component compatible with Java 11
```
Because this version is not a release, and still we have one version ahead (7.4), we should still build with Java11 for current users.  
Wondering if we could set a multiple variant attributes for the different JVM in the artifact. 

## Java17 integration

At this point, we need to support builds with different environments, tests using AGP 8.0.0 and JDK11 returns:
```
> Android Gradle plugin requires Java 17 to run. You are currently using Java 11. |  
-- | --
  | Your current JDK is located in /Library/Java/JavaVirtualMachines/zulu-11.jdk/Contents/Home |  
  | You can try some of the following options: |  
  | - changing the IDE settings. |  
  | - changing the JAVA_HOME environment variable. |  
  | - changing `org.gradle.java.home` in `gradle.properties`.
```

For that reason we separated the Java17 configuration from the current one. This changed required:
* Support in the tasks and versioning for multiple java version files 
* Create new job to execute the AGP versions with JDK 17

### Java17 supported versions file
Created new version file including the Java 17 supported versions. I included only the new AGP version, but I would be more ambitious including at least the latest stable AGP version. 

This file had to be included in the parsers that represents:
*  Test task for each Android version and run the tests annotated with the MultiVersionTest category 
*  Supported versions of Android and Gradle versions. Used later in tests

### CI 
New job `android_version_tests_java_17` executing integration tests in a Java17 container provided by the new versions file.
Screen Shot 2023-01-04 at 7 28 46 PM

New matrix versions representing Java17 supported versions, this new matrix allows to continue extending future versions.

During the execution, we noticed that the Gradle Runner build tests were executed with Java11. We understood this was caused by the toolchain configuration of the project. We verified that the JAVA_HOME of the container was 17. but the JAVA_HOME during runtime was 11. For this reason we've created a new environment variable with the content of the JAVA_HOME of the container. Later we use this variable to set the project jvm value.

### Additional Notes
We need to update the process of AGPs including the new versioning file when updating AGPs. 

## Results
AGP 8 with Java17: https://ge.solutions-team.gradle.com/s/f3aocdtemg55s 
AGP 7.4 with Java11: https://ge.solutions-team.gradle.com/s/itirikmxni4a2


runningcode:
I propose we close this in favor of #406 :)

PR #406, opened 2023-01-05
Update to AGP 8.0.0-alpha11

runningcode:
This sets the tests run using AGP 8.0.0 or higher to use JDK 19.


cdsap:
I prefer your approach! More simpler and ignoring the complexity of CI. 
Regarding JDK19, what about 17? I think we need to cover the 17 because is the major change on the new agp.

cdsap (approved):

Issue #412, opened 2023-01-13
Issue with AGP 7.4.0


eboudrant:
Hello, I'm trying AGP 7.4.0 using latest version and it fails to apply the plugin : 

```
Caused by: java.lang.NullPointerException: Cannot invoke method selector() on null object
	at org.gradle.android.workarounds.RoomSchemaLocationWorkaround.applyToAllAndroidVariantsWithNewVariantApi(RoomSchemaLocationWorkaround.groovy:277)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at org.gradle.android.workarounds.RoomSchemaLocationWorkaround.applyToAllAndroidVariants(RoomSchemaLocationWorkaround.groovy:257)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at org.gradle.android.workarounds.RoomSchemaLocationWorkaround.apply(RoomSchemaLocationWorkaround.groovy:81)
	at org.gradle.android.AndroidCacheFixPlugin$_apply_closure1.doCall(AndroidCacheFixPlugin.groovy:69)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at org.gradle.android.AndroidCacheFixPlugin.apply(AndroidCacheFixPlugin.groovy:67)
	at org.gradle.android.AndroidCacheFixPlugin.apply(AndroidCacheFixPlugin.groovy)
[...]

PR #414, opened 2023-01-13
Apply configuration new variants API only when Plugins are present

cdsap:
Issue #412 describes the issue when using AGP 7.4 and the Plugin throws:
```
Failed to apply plugin 'org.gradle.android.cache-fix'.
> Cannot invoke method selector() on null object
```
This is caused when the plugin is configuring a project without Android plugins applied. 
I can reproduce the error when in the project `nowinandroid` I remove the `apply false` in the plugin definition. That means the rootProject is trying to retrieve the `androidComponents`:
```
 def androidComponents = project.extensions.findByName("androidComponents")
```
when they are not applied. This is a change in the AGP not reproducible in 7.3.1. 

Similar to the old implementation this PR only applies the configuration when a project contains the Android plugins.

### DIFF
--- a/src/main/groovy/org/gradle/android/workarounds/room/androidvariants/ApplyAndroidVariants.groovy
+++ b/src/main/groovy/org/gradle/android/workarounds/room/androidvariants/ApplyAndroidVariants.groovy
@@ -34,6 +34,16 @@ class ApplyAndroidVariants {
     }
 
     private static void applyToAllAndroidVariantsWithNewVariantApi(Project project, ConfigureVariants configureVariants) {
+        project.plugins.withId("com.android.application") {
+            configureNewVariants(project, configureVariants)
+        }
+
+        project.plugins.withId("com.android.library") {
+            configureNewVariants(project, configureVariants)
+        }
+    }
+
+    private static configureNewVariants(Project project, ConfigureVariants configureVariants){
         def androidComponents = project.extensions.findByName("androidComponents")
         def selector = androidComponents.selector()
         androidComponents.onVariants(selector.all(), configureVariants.newVariantConfiguration)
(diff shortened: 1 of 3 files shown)


eboudrant:
I looked also and noticed we apply the plugin using this in our root `build.gradle`: 
```
subprojects { subProject ->
    apply plugin: "org.gradle.android.cache-fix"
```
The error probably happen because we apply it regardless it's a project with or without android plugin. We should probably fix our build script too, thanks for looking quickly at the issue!

runningcode (approved):
let's add a test for this as well.

cdsap:
@runningcode added unit tests covering plugin block configuration with/wo applying the plugin by default

src/main/groovy/org/gradle/android/workarounds/room/androidvariants/ApplyAndroidVariants.groovy
```diff
@@ -34,6 +34,16 @@ class ApplyAndroidVariants {
     }
 
     private static void applyToAllAndroidVariantsWithNewVariantApi(Project project, ConfigureVariants configureVariants) {
+        project.plugins.withId("com.android.application") {
+            configureNewVariants(project, configureVariants)
+        }
+
+        project.plugins.withId("com.android.library") {
```
daio, line 41: It's already merged, but shouldn't the plugin also apply to modules that use com.android.test plugin?

eboudrant: I think it should also apply to modules with `com.android.dynamic-feature` plugin.

cdsap: @daio @eboudrant you're right. 
Checking the hierarchy of the plugins:
Screen Shot 2023-01-15 at 10 38 40 AM
I'm wondering if checking by `com.android.base` would be enough. Next week will work on this

Issue #416, opened 2023-01-14
Cache miss with AGP 7.4


GianfrancoMS:
At my company, we're trying to update to AGP 7.4 and to the Android Cache Fix Gradle Plugin 2.6.2, but we have encountered a cache miss with the task `compile${variant}JavaWithJavac`. This cache miss did not happen in previous versions.

Versions:
- Gradle 7.5.1
- Kotlin 1.7.21
- AGP 7.4 (previous was 7.3.1)
- Android Cache Fix Gradle Plugin 2.6.2 (previous was 2.6.1)
- JDK 17

<img width="1216" alt="Screenshot 2023-01-14 at 11 55 20" src="https://user-images.githubusercontent.com/22310159/212485107-79a4a7e1-ee4a-42e2-bd7e-86a00fcab503.png">

Let me know if you need a scan dump since my GE instance is behind a VPN.

cdsap:
hi @GianfrancoMS, thanks for reporting this issue. 
I notice the issue is coming from the `jdkImage`. We provided a JdkImageWorkaround for AGP < 7.4 but the workaround is not applied for AGP 7.4 since the fix was provided in AGP 7.4.0-alpha07 (https://issuetracker.google.com/u/1/issues/234820480#comment6)

If you test 2.6.2 with AGP 7.3.1 works right?

We will test again the AGP 7.4 reproducing the JdkImage issue. If is not fixed, we will apply again the workaround in the latest AGP.

GianfrancoMS:
2.6.2 with AGP 7.3.1 works perfectly. It only happens with AGP 7.4.

cdsap:
Hi @GianfrancoMS, I finished the investigation and here is the result:

The existing issue mentioned https://issuetracker.google.com/u/1/issues/234820480 is not reproducible in AGP 7.4, so the workaround is correctly not applied in the AGP 7.4.

Now, why are we seeing this cache miss in the JavaCompile? 
First, I reproduced your issue in a simple project with AGP 7.4 and Android Cache Fix Plugin 2.6.2 executing builds in different Java versions:
![Screen Shot 2023-01-17 at 6 28 12 PM](https://user-images.githubusercontent.com/475733/213067020-b7129124-c998-4294-acd0-84ee24b7fcad.png)
I noticed that the reason is different from one project using 7.3.1:
[...]

Issue #341, opened 2022-09-01
ignore '**/java/lang/invoke/**' from runtimeClassPath


kenyee:
We're hitting transform differences in cache artifacts:
```
Only in /tmp/extracted/java.base: NOTICES
Binary files extracted/java.base/java/lang/invoke/DelegatingMethodHandle$Holder.class and /tmp/extracted/java.base/java/lang/invoke/DelegatingMethodHandle$Holder.class differ
Binary files extracted/java.base/java/lang/invoke/DirectMethodHandle$Holder.class and /tmp/extracted/java.base/java/lang/invoke/DirectMethodHandle$Holder.class differ
```
I was asking about it in this thread: https://gradle-community.slack.com/archives/CAHSN3LDN/p1661952779052469

And someone mentioned, they added this to work around this issue:
```
allProjects {
  normalization {
    runtimeClasspath {
      ignore '**/module-descriptor.txt' // don't need if you are up to date on AGP cache plug in fix
      ignore '**/java/lang/invoke/**'
    }
}
```

It looks like this plugin has a JdkImageWorkaround class that does this normalization as well but doesn't cover this case:
https://github.com/gradle/android-cache-fix-gradle-plugin/blob/d7f486e318e3698afa58834917b781cb0ef151c8/src/main/groovy/org/gradle/android/workarounds/JdkImageWorkaround.groovy

Would be nice if it were done in this plugin as well so we don't have to and this would improve cache hits for everyone.

kenyee:
It's an internal Twitter OpenJDK11 build unfortunately (our folks tweak it for server usage and then we all use it).
I'll check w/ our VM folks to confirm it's based on the same source code and version.

kenyee:
FWIW, our VM folks confirmed same version.

That $Holder class is apparently generated code: https://github.com/openjdk/jdk11u/blob/73eef16128417f4a489c4dde47383bb4a00f39d4/src/jdk.jlink/share/classes/jdk/tools/jlink/internal/plugins/GenerateJLIClassesPlugin.java#L433
https://github.com/openjdk/jdk11u/blob/73eef16128417f4a489c4dde47383bb4a00f39d4/src/java.base/share/classes/java/lang/invoke/GenerateJLIClassesHelper.java#L105

And the code varies slightly between x86/ARM architectures despite it generating JVM bytecode.  That's pretty gnarly code...never knew the JVM did this... 😸 
[...]

PR #419, opened 2023-01-18
Including JdkImageWorkaround in newer versions

cdsap:
This PR includes the `JdkImageWorkaround` in AGP >= 7.4 versions again. 

## Context
After issue #416 we investigated the state of the cache miss for Java compilation in recent AGP 7.4.  The complete investigation is here https://github.com/gradle/android-cache-fix-gradle-plugin/issues/416#issuecomment-1386398170
The investigation was based on a scenario with two different java versions: Temurin 11 latest and Zulu 11 latest.
Today, we extended the reproducible project with Zulu 11.0.9+1 with the following results:
| Version Variant A | Version Variant B | Cache | Build Scan |
|-------------------|-------------------|---------|-----|
| Zulu 11 latest    | Temurin 11 latest | Hit   |  https://ge.solutions-team.gradle.com/c/zx27tbs54ehry/3vkahm557urb4/task-inputs |
| Zulu 11 latest    | Zulu 11.0.9+1     | Miss  |  https://ge.solutions-team.gradle.com/c/pobsvmnz6aywq/3vkahm557urb4/task-inputs |
| Temurin 11 latest | Zulu 11.0.9+1     | Miss  | https://ge.solutions-team.gradle.com/c/pobsvmnz6aywq/zx27tbs54ehry/task-inputs |

Sadly we observed cache misses using the Java version Zulu 11.0.9+1. 

Additionally, we observed same issue on local builds using an M1 with Java versions:  liberica-11.0.15+10 and liberica-11.0.15.1+2: https://ge.solutions-team.gradle.com/c/mefzwvhujyq4y/fgdorbiinpkfw/task-inputs

The main issue is that the output difference is non-deterministic. With some Java combinations, the issue can be fixed by applying normalization(https://github.com/gradle/android-cache-fix-gradle-plugin/issues/416#issuecomment-1396216254), but in others, we need to apply the workaround.

## Implementation
We removed the annotation property `fixed` in the workaround. To make it work with AGP 7.4 we had to:
* Check the current AGP version because `variant.javaCompileProvider` is not available in the new variant API. Instead of that, we configure the `JavaCompile` task
* Following the same issue of Room(#412), we needed to apply correctly the build variants due to the change in the AGP implementation. In this case, we are applying the workaround based on `android.base` plugin.
* Once the workaround was applied, we noticed for different scenarios differences on the `java/lang/invoke` package. I included `it.ignore '**/java/lang/invoke/**'` in the workaround normalization. I will check with the team.

## Results
M1 Local build using liberica-11.0.15+10 and liberica-11.0.15.1+2  --> https://ge.solutions-team.gradle.com/c/7g6lcqlsv6qje/lbdwlkkg4iyyg/task-inputs


src/main/groovy/org/gradle/android/workarounds/JdkImageWorkaround.groovy
```diff
@@ -86,14 +98,17 @@ class JdkImageWorkaround implements Workaround {
...
+            def androidComponents = project.extensions.findByName("androidComponents")
+            def selector = androidComponents.selector()
+            androidComponents.onVariants(selector.all(), configureVariant)
+        }
     }
 
     static def applyRuntimeClasspathNormalization(Project project) {
         project.normalization { handler ->
             handler.runtimeClasspath {
+                it.ignore '**/java/lang/invoke/**'
```
cdsap, line 111: @runningcode I think we need to include this normalization in the workaround.
I know it was discussed here https://github.com/gradle/android-cache-fix-gradle-plugin/issues/341#issuecomment-1234700697 but I wanted to confirm.

In a M1 scenario with AGP 7.4 and different java versions, if we remove this line, the result is:
https://ge.solutions-team.gradle.com/c/gv6iro7iyouvy/xituylgavkwve/task-inputs?expanded=WyJqZ3BseXgydm90d3dpLW9wdGlvbnMuY29tcGlsZXJhcmd1bWVudHByb3ZpZGVycy4kMy5leHRyYWN0ZWRqZGtpbWFnZSJd
We need to apply the normalization manually to fix this issue. So the workaround is not enough(Similar to 7.3.1)
Wdyt?

runningcode: Happy to discuss this again. Even though it is happening in 7.4 just like it was happening in 7.3. I still stand by my comment there and I suggest removing this line for the same reasons. I don't think we should ignore `java/lang/invoke` implicitly across all modules where this plugin is applied.

src/main/groovy/org/gradle/android/workarounds/JdkImageWorkaround.groovy
```diff
@@ -53,16 +53,28 @@ class JdkImageWorkaround implements Workaround {
...
-                def jdkImageInput = getJdkImageInput(task)
-                if (jdkImageInput != null) {
-                    setupExtractedJdkImageInputTransform(project, getJvmHome(task))
-                    replaceCommandLineProvider(task, jdkImageInput)
+            if (Versions.CURRENT_ANDROID_VERSION <= VersionNumber.parse("7.4.0-alpha01")) {
+                variant.javaCompileProvider.configure { JavaCompile task ->
+                    jdkTransform(project, task)
+                }
+            } else {
+                project.afterEvaluate {
```
runningcode, line 61: why do we need `afterEvaluate` here?

src/main/groovy/org/gradle/android/workarounds/JdkImageWorkaround.groovy
```diff
@@ -53,16 +53,28 @@ class JdkImageWorkaround implements Workaround {
         applyRuntimeClasspathNormalization(project)
 
         applyToAllAndroidVariants(project) { variant ->
-            variant.javaCompileProvider.configure { JavaCompile task ->
-                def jdkImageInput = getJdkImageInput(task)
-                if (jdkImageInput != null) {
-                    setupExtractedJdkImageInputTransform(project, getJvmHome(task))
-                    replaceCommandLineProvider(task, jdkImageInput)
+            if (Versions.CURRENT_ANDROID_VERSION <= VersionNumber.parse("7.4.0-alpha01")) {
```
runningcode, line 56: is there a newer replacement for `variant.javaCompileProvider` in 7.4 + ?

runningcode (commented):
Thanks for the build scan comparisons and the detailed PR description.

src/main/groovy/org/gradle/android/workarounds/JdkImageWorkaround.groovy
cdsap: here correct me if I'm wrong, but my approach was based in:
because the `variant.javaCompilerProvider` is unavailable, I need to check if a compiler argument is defined for jdkImage. When I'm configuring the project, the AGP didn't create the compiler argument yet. That's the main reason. Ideas/suggestions?

src/main/groovy/org/gradle/android/workarounds/JdkImageWorkaround.groovy
cdsap: sadly nope, https://androidstudygroup.slack.com/archives/C03KKLGC2/p1591411391491800

src/main/groovy/org/gradle/android/workarounds/JdkImageWorkaround.groovy
cdsap: My only preference is to avoid the normalization from the client side, other than that I'm ok removing it...

src/main/groovy/org/gradle/android/workarounds/JdkImageWorkaround.groovy
runningcode: Can we do something like `variant.javaCompilation` like we do in the `RoomWorkaround`?

https://github.com/gradle/android-cache-fix-gradle-plugin/blob/main/src/main/groovy/org/gradle/android/workarounds/RoomSchemaLocationWorkaround.groovy#L116

src/main/groovy/org/gradle/android/workarounds/JdkImageWorkaround.groovy
runningcode: See above thread ^

src/main/groovy/org/gradle/android/workarounds/JdkImageWorkaround.groovy
runningcode: I see both sides but I think normalizing such an important package like `java/lang/invoke` should be done explicitly. Unfortnately, normalization applies to all tasks in the project and not just the one task we care about here.

src/main/groovy/org/gradle/android/workarounds/JdkImageWorkaround.groovy
cdsap: tried yesterday and same issue, at the time the variant is being processed we don't have agp `jdkImageInput` required to apply the workaround.

src/main/groovy/org/gradle/android/workarounds/JdkImageWorkaround.groovy
cdsap: removed https://github.com/gradle/android-cache-fix-gradle-plugin/pull/419/commits/cf863228d1e0276c3ee94de9d9a7fbdae6759f94

src/main/groovy/org/gradle/android/workarounds/JdkImageWorkaround.groovy
cdsap: to extend the previous comment, in the previous AGP variant you have access to the Task Provider of Java task, because is lazy we can guarantee that the required ap options have been configured. 
The new variant removes this functionality, in the case of the RoomWorkaround we don't require to have existing ap options because we are just inserting new argument provider in the variant block.

cdsap: @runningcode, following Xavier's comments, there is no replacement for `variant.javaCompileProvider` but we can still use the old API before is removed in AGP 9. 
I created new commit using the old API https://github.com/gradle/android-cache-fix-gradle-plugin/pull/419/commits/fa6672849301f8e69a3fbcb1e5853d341f8a98db and worked

src/main/groovy/org/gradle/android/workarounds/JdkImageWorkaround.groovy
```diff
@@ -54,24 +54,22 @@ class JdkImageWorkaround implements Workaround {
 
         applyToAllAndroidVariants(project) { variant ->
             variant.javaCompileProvider.configure { JavaCompile task ->
-                def jdkImageInput = getJdkImageInput(task)
-                if (jdkImageInput != null) {
-                    setupExtractedJdkImageInputTransform(project, getJvmHome(task))
-                    replaceCommandLineProvider(task, jdkImageInput)
-                }
+                jdkTransform(project, task)
```
runningcode, line 57: can we keep the code exactly as before to minimize the diff?

src/main/groovy/org/gradle/android/workarounds/JdkImageWorkaround.groovy
runningcode: That sounds good. Thanks for getting Xav's feedback there.

src/main/groovy/org/gradle/android/workarounds/JdkImageWorkaround.groovy
cdsap: done

runningcode (approved):
Nice code cleanup here.

PR #420, opened 2023-01-18
Using base plugin for Room workaround

cdsap:
Based on this discussion https://github.com/gradle/android-cache-fix-gradle-plugin/pull/414#discussion_r1070514914
We configure the AGP 7.4 variants with `project.plugins.withId("com.android.base") {` instead of checking for specific plugins like application/library.


runningcode:
Is it common to apply room to these other types of modules with the dynamic feature plugin and the testing plugin?

runningcode (approved):

cdsap:
@runningcode I don't think so

runningcode:
Looks like a good change to simpify the codebase either way.

Issue #432, opened 2023-02-10
JavaCompileWorkaround throws 'Cannot query the value of this property because it has no value available' on v2.6.4


tomkoptel:
Hello,

I am experiencing the following issue after upgrading to `v2.6.4`. I am using Kotlin `1.7.22`, Android Gradle Plugin `7.4.1` and Gradle `7.6`.

The task `./gradlew :app:lintRelease --stacktrace --no-configuration-cache`

```
> Configure project :build-logic:android
WARNING: Unsupported Kotlin plugin version.
The `embedded-kotlin` and `kotlin-dsl` plugins rely on features of Kotlin `1.7.10` that might work differently than in the requested version `1.7.22`.

> Configure project :build-logic:jacoco
WARNING: Unsupported Kotlin plugin version.
The `embedded-kotlin` and `kotlin-dsl` plugins rely on features of Kotlin `1.7.10` that might work differently than in the requested version `1.7.22`.
Is CI env? Answer: null

> Configure project :app
Warning: The 'kotlin-android-extensions' Gradle plugin is deprecated. Please use this migration guide (https://goo.gle/kotlin-android-extensions-deprecation) to start working with View Binding (https://developer.android.com/topic/libraries/view-binding) and the 'kotlin-parcelize' plugin.
AndroidMetadataPlugin: Found all the info necessary to sign the artifact
DynamicLocalePlugin: 'available_languages' key is not set applying 'defaultConfig { resConfigs([en]) }'.

> Configure project :dynamic-locale

> Task :dynamic-locale:compileReleaseJavaWithJavac FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':dynamic-locale:compileReleaseJavaWithJavac'.
> Cannot query the value of this property because it has no value available.

* Try:
> Run with --info or --debug option to get more log output.

* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':dynamic-locale:compileReleaseJavaWithJavac'.
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.lambda$executeIfValid$1(ExecuteActionsTaskExecuter.java:142)
        at org.gradle.internal.Try$Failure.ifSuccessfulOrElse(Try.java:282)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeIfValid(ExecuteActionsTaskExecuter.java:140)
[...]

PR #433, opened 2023-02-10
Check if schema exists to avoid JavaCompileWorkaround issue

cdsap:
After workaround extractions and 2.6.4, users reported #432 
This issue is reproducible when applying the plugin in modules without kapt or room implementation.

The extraction didn't verify if the schema was present like in version 2.6.3: https://github.com/gradle/android-cache-fix-gradle-plugin/blob/v2.6.3/src/main/groovy/org/gradle/android/workarounds/RoomSchemaLocationWorkaround.groovy#L259

This PR adds the check. Kapt workaround is not affected. 
Internal note: Why we don't have tests covering builds with this configuration?


runningcode (approved):

StylianosGakis:
Thanks for the super quick fix and release 😊

PR #429, opened 2023-02-07
Ksp Workaround

cdsap:
Applies KspRoomWoraround. 

# Implementation
Created new `KspWorkaround` and `KspRoomSchemaLocationArgumentProvider` to generate the schemas to a temporary directory,  then copy the generated schemas to the registered output directory(similar to kapt)

Ksp uses a different format for the compiler argument:
```
if (this instanceof KspRoomSchemaLocationArgumentProvider) {
    return ["${RoomSchemaLocationWorkaround.ROOM_SCHEMA_LOCATION}=${schemaLocationPath}" as String]
} else {
    return ["-A${RoomSchemaLocationWorkaround.ROOM_SCHEMA_LOCATION}=${schemaLocationPath}" as String]
}
```

# Ksp Workaround Tests 
Separated Ksp tests in a new Test Class. Created `RoomWorkaroundAbstractTest` to encapsulate the common logic. Ksp tests support `["1.7.22-1.0.8", "1.8.10-1.0.9"]`.  We found an incompatibility with Kotlin 1.8.0 with Ksp 1.7.22(this combination is skipped)

PR #438, opened 2023-02-15
Updating parameters used in gradle properties for the Kotlin daemon Simple app

cdsap:
During the CI builds of the KspWorkaround(#429), we noticed builds failing in the tests executed:
Screenshot 2023-02-14 at 5 46 15 PM
Test failing showed flakiness in some of the executions:
Screenshot 2023-02-14 at 5 48 16 PM
Logs showed:
```
Gradle build daemon disappeared unexpectedly (it may have been killed or may have crashed)
```
This error happens at the level of the test, not in the Build containing the Gradle runner builds. 

At this point, we started monitoring more in detail the number of test executed for a given build: 
For instance, for AGP 7.3.1 we execute around 40 unit tests, some of those tests execute different builds using different Kotlin versions. This combination causes to run more than 55 builds per build invocation.  We enabled process tracking for the Kotlin processes and discovered that 6 Kotlin processes were created:
Screenshot 2023-02-14 at 5 57 13 PM
We are creating one Kotlin process for each Kotlin version, additionally some of the tests use different Gradle versions, explaining why we have 6 Kotlin processes.

The resources at the GHA container are limited 7GB, meaning that the container is killing the Gradle daemons because the pressure of the different processes. After some research, we ended up configuring the parameters of the Kotlin daemon limiting the heap memory allocation.
This PR includes the new configuration assuming that the max limit of Kotlin processes is 6 and satisfying the next formula:
  
```
(Kotlin Process * 6) + (Gradle Daemon * 2) +(Gradle Daemon(container)) + Test process < 7 GB
```

With this approach, we are reducing the number of total builds to 380 per commit. Before, we were incresing up to 524 builds because the retry, additionally the flakyness dissapear.

### DIFF
--- a/src/test/groovy/org/gradle/android/SimpleAndroidApp.groovy
+++ b/src/test/groovy/org/gradle/android/SimpleAndroidApp.groovy
@@ -138,7 +138,7 @@ class SimpleAndroidApp {
 
         file("gradle.properties") << """
                 android.useAndroidX=true
-                org.gradle.jvmargs=-Xmx1536m -XX:+UseParallelGC -Dkotlin.daemon.jvm.options=-Xmx768m,-Xms256m,-XX:MaxMetaspaceSize=1g
+                org.gradle.jvmargs=-Xmx1536m -Dkotlin.daemon.jvm.options=-Xmx768m,-Xms256m
                 kapt.use.worker.api=${kaptWorkersEnabled}
                 android.experimental.enableSourceSetPathsMap=true
                 android.experimental.cacheCompileLibResources=true
--- a/src/test/groovy/org/gradle/android/SimpleAndroidApp.groovy
+++ b/src/test/groovy/org/gradle/android/SimpleAndroidApp.groovy
@@ -138,7 +138,7 @@ class SimpleAndroidApp {
 
         file("gradle.properties") << """
                 android.useAndroidX=true
-                org.gradle.jvmargs=-Xmx2048m -Dkotlin.daemon.jvm.options=-Xmx1500m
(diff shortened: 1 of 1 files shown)


src/test/groovy/org/gradle/android/SimpleAndroidApp.groovy
```diff
@@ -138,7 +138,7 @@ class SimpleAndroidApp {
 
         file("gradle.properties") << """
                 android.useAndroidX=true
-                org.gradle.jvmargs=-Xmx2048m -Dkotlin.daemon.jvm.options=-Xmx1500m
+                org.gradle.jvmargs=-Xmx1536m -XX:+UseParallelGC -Dkotlin.daemon.jvm.options=-Xmx768m,-Xms256m,-XX:MaxMetaspaceSize=1g
```
runningcode, line 141: does it make sense that the maxmetaspacesize is greater than the heap size?

also regarding parallelGC, it might be more performant but it does not recliam memory as quickly as G1. does switching to parallel give other benefits opposed to G1?

runningcode (commented):
Thanks for investigating this!

src/test/groovy/org/gradle/android/SimpleAndroidApp.groovy
cdsap: done https://github.com/gradle/android-cache-fix-gradle-plugin/pull/438/commits/005cccfe95885a270caf6eba977755fe2f201a2e

didn't observe any issues in the CI build, asking for review again

runningcode (approved):
I like the minimal set of changes

PR #440, opened 2023-02-16
Use temurin instead of adopt java distribution

c00ler:
As per note in the https://github.com/actions/setup-java#supported-distributions section adopt openjdk was moved to eclipse temurin and it is advised to update


runningcode (approved):
Thanks Alexey!

PR #450, opened 2023-02-24
Enable PTS

tylerbertrand:
Conditionally enable PTS for all tests if the project property 'isPTSEnabled' evaluates to true

Enable PTS for all tests in CI if not running against main


runningcode (approved):
Thanks, looks good to me!

PR #457, opened 2023-03-01
Enable PTS

tylerbertrand:
Conditionally enable PTS if the project property 'isPTSEnabled' evaluates to true

Enable PTS in CI if not running against main

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -107,6 +107,9 @@ tasks.withType(Test).configureEach {
         maxRetries = isCI ? 1 : 0
         maxFailures = 20
     }
+    predictiveSelection {
+        enabled = providers.gradleProperty("isPTSEnabled").map { it != 'false' } .orElse(false)
+    }
 }
 
 // Generate a test task for each Android version and run the tests annotated with the MultiVersionTest category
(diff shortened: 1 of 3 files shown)


runningcode (approved):

runningcode (changes requested):
if the property is present, but not set to true, we should also enable PTS. How does it behave if the property is present but not set?

build.gradle
```diff
@@ -107,6 +107,9 @@ tasks.withType(Test).configureEach {
         maxRetries = isCI ? 1 : 0
         maxFailures = 20
     }
+    predictiveSelection {
+        enabled = providers.gradleProperty("isPTSEnabled").map({ it.toBoolean() }).orElse(false)
```
erichaagdev, line 111: You can drop the parenthesis for `map`.

```suggestion
        enabled = providers.gradleProperty("isPTSEnabled").map { it.toBoolean() } .orElse(false)
```

build.gradle
```diff
@@ -107,6 +107,9 @@ tasks.withType(Test).configureEach {
         maxRetries = isCI ? 1 : 0
         maxFailures = 20
     }
+    predictiveSelection {
+        enabled = providers.gradleProperty("isPTSEnabled").map { it != 'false' } .orElse(false)
```
erichaagdev, line 111: Minor formatting suggestion on my previously suggested minor formatting suggestion:

```suggestion
        enabled = providers.gradleProperty("isPTSEnabled").map { it != "false" }.orElse(false)
```

erichaagdev (approved):

PR #461, opened 2023-03-02
Defer plugin applying in subprojects

Goooler:


runningcode:
What does this do and what is the benefit?

Goooler:
https://docs.gradle.org/current/userguide/task_configuration_avoidance.html
https://github.com/liutikas/gradle-best-practices#avoid-the-all-callback-on-gradles-container-types

runningcode:
Thanks for the pointers. These suggestions apply to tasks, not to plugins. I don't see a benefit here.

bddckr:
I believe it still applies because [the documentation of the used `DomainObjectCollection.withType` method overload](https://github.com/gradle/gradle/blob/b73ac0594eb570ecf8f6d43727499f7f23cce33c/subprojects/core-api/src/main/java/org/gradle/api/DomainObjectCollection.java#L67-L75) says that it's the same as calling `all`, [which isn't avoiding configuration according to the first provided link from Goooler above](https://github.com/gradle/gradle/blob/b73ac0594eb570ecf8f6d43727499f7f23cce33c/subprojects/docs/src/docs/userguide/extending-gradle/task_configuration_avoidance.adoc?plain=1#L453-L455).

erichaagdev:
Hey @Goooler and @bddckr! Thank you for your interest in improving our documentation. 

[Task configuration avoidance](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html) only applies to tasks and the `TaskContainer`. It does not apply to plugins or the `PluginContainer`, or any other container for that matter.

Issue #446, opened 2023-02-18
RoomSchemaLocationArgumentProvider added to Kapt modules


cdsap:
When applying the Android Cache Fix Gradle plugin to a project using `kapt` the `RoomSchemaLocationArgumentProvider` is created always, even if the project is not using `room-compiler`.

# Steps to reproduce
1- Apply the plugin in one project using kapt without `room-compiler`
2- Add in the root build gradle file:
```
    tasks.withType<org.jetbrains.kotlin.gradle.internal.KaptWithoutKotlincTask>().configureEach {
        doLast {
            this@configureEach.annotationProcessorOptionProviders.forEach {
                println(it.toString())
            }
        }
    }
```
Result:
```
> Task :core:common:kaptDemoDebugKotlin
[dagger.hilt.android.plugin.HiltGradlePlugin$configureProcessorFlags$1$1@45020498, 
org.gradle.android.workarounds.RoomSchemaLocationWorkaround$KaptRoomSchemaLocationArgumentProvider@361e7a4b]
```

I tested with older versions (2.5.7, Sep 2022) and it's reproducible

runningcode:
This was intentional. At the time there was no way around this without eagerly resolving dependencies. Do you see how this can be avoided?

cdsap:
Not sure if it's possible. Ideally, we would check in which processor has been applied the room-processor and then avoid the unused configuration on kapt modules

runningcode:
Let's not do this and instead favor Google's Room Plugin.

PR #470, opened 2023-03-08
Verify room-processor is included before adding the ArgumentProvider

cdsap:
Issue #446 
Kapt workaround always adds the argument provider for the variants of the module using kapt plugin independently if they are using the room processor. 
This PR adds the check to avoid the configuration of the argument provider if `room-processor` is not set.

### DIFF
--- a/src/main/groovy/org/gradle/android/workarounds/room/KaptWorkaround.groovy
+++ b/src/main/groovy/org/gradle/android/workarounds/room/KaptWorkaround.groovy
@@ -39,18 +39,22 @@ class KaptWorkaround extends AnnotationProcessorWorkaround<KaptRoomSchemaLocatio
             @Override
             Closure<?> getOldVariantConfiguration() {
                 return { variant ->
-                    def variantSpecificSchemaDir = project.objects.directoryProperty()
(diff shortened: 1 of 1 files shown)


src/main/groovy/org/gradle/android/workarounds/room/KaptWorkaround.groovy
```diff
@@ -143,4 +147,11 @@ class KaptWorkaround extends AnnotationProcessorWorkaround<KaptRoomSchemaLocatio
             throw new RuntimeException("Field '${fieldName}' not found")
         }
     }
+
+    // Before adding the ArgumentProvider, we need to verify that the room-processor is applied in the kapt configuration
+    private boolean kaptContainsRoom() {
+        return project.configurations.named("kapt").get().dependencies.name
```
cdsap, line 153: @bigdaz can you confirm this is correct? 

At this point, we are checking the configurations after the android variants have been resolved.

cdsap: thanks @bigdaz  +1000

cdsap:
After testing @bigdaz points, I verified that using:
```
val someConfiguration by configurations.creating
configurations {
  
    someConfiguration
    configurations.getByName("kapt") {
        extendsFrom(someConfiguration)
    }
}

```
doesn't catch the room compiler, closing until new approach

Issue #460, opened 2023-03-02
Increase Integration Tests coverage


cdsap:
Most of the tests use a simple project where we configure the same scenario:
* Default configuration ( apply false for our plugin) 
* Unique processor is applied in the test (Kapt/Apt/Ksp)
* Ksp is applied in the root build classpath

However, we didn't catch issues like https://github.com/gradle/android-cache-fix-gradle-plugin/issues/412. We need to increase the coverage for the different scenarios like:
* Multiple processors included in the same module 
* Apply the plugin in non-Android projects
* Modules using kapt/ksp without the room-compiler
* Ksp is defined at the module level instead of the main build classpath.
* ...

PR #477, opened 2023-03-21
Extracting source/resources required for the GradleRunner tests

cdsap:
SimpleAndroid class is getting complex, and with more use cases to cover the class is getting hard to maintain. 
Before starting the rework of the class(#460) we need to clean up the current class.
This PR extracts classes/resources used in the "Simple" projects run under the Gradle Runner tests


src/test/groovy/org/gradle/android/writer/XmlResources.groovy
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    static String emptyManifest() {
+        return '''<?xml version="1.0" encoding="utf-8"?>
+                <manifest xmlns:android="http://schemas.android.com/apk/res/android">
+                </manifest>
+        '''.stripIndent()
+    }
+
+    static String strings() {
```
jprinet, line 50: I find this method name a bit confusing

jprinet (approved):
Your version is way more readable 👍

src/test/groovy/org/gradle/android/writer/Kotlin.groovy
```diff
@@ -0,0 +1,13 @@
+package org.gradle.android.writer
+
+class Kotlin {
```
runningcode, line 3: I dislike the name `Kotlin` as a class name. I'm thinking it might be better to group these code writing classes to something like `CodeSnippets`. Then we can call `CodeSnippets.getKotlinDataClass()` and `CodeSnippets.getJavaClass()`. WDYT?

src/test/groovy/org/gradle/android/writer/XmlResources.groovy
cdsap: done, changed to getXmlStrings

src/test/groovy/org/gradle/android/writer/Kotlin.groovy
cdsap: sounds good. 
Grouped all sources/resources in`CodeSnippets` https://github.com/gradle/android-cache-fix-gradle-plugin/pull/477/commits/c101f757302e52ceac73366942d47fbfb7fe0ed6

src/test/groovy/org/gradle/android/SimpleAndroidApp.groovy
```diff
@@ -1,5 +1,6 @@
 package org.gradle.android
 
+
```
jprinet, line 3: empty blank line

jprinet (approved):
LGTM

PR #481, opened 2023-04-04
Use relative path on `room.schemaLocation` annotation processor argument

gmazzo:
Researching some cache misses we found our that `ksp` tasks are not relocatable (miss if build on another absolute path) when using `room` compiler.

Further research comparing scans revealed that the cache mismatch happens because there is a difference in the `pluginOptions.$0.asTaskInputArgs` input property:
image

Digging deeper on the property, it's revealed that `RoomSchemaLocationArgumentProvider` is adding absolute paths to the schema location direction in the annotation processor params:
```
pluginOptions.$0.asTaskInputArgs: {com.google.devtools.ksp.symbol-processing.incremental=true, com.google.devtools.ksp.symbol-processing.incrementalLog=false, com.google.devtools.ksp.symbol-processing.allWarningsAsErrors=false, com.google.devtools.ksp.symbol-processing.returnOkOnError=true, com.google.devtools.ksp.symbol-processing.apoption.0=glovo.featureToggles.enumName=com.glovoapp.payments.PaymentsFeatureToggles, com.google.devtools.ksp.symbol-processing.apoption.1=glovo.featureToggles.includeDependencies=false, com.google.devtools.ksp.symbol-processing.apoption.2=glovo.featureToggles.hiltSupport=true, com.google.devtools.ksp.symbol-processing.apoption.3=room.schemaLocation=/Users/gmazzola/Documents/glovo-customer-android/payments/build/roomSchemas/kspDebugKotlinTemp, com.google.devtools.ksp.symbol-processing.excludedProcessors=, com.google.devtools.ksp.symbol-processing.mapAnnotationArgumentsInJava=false}
```

Looking at `com.google.devtools.ksp.symbol-processing.apoption.3=room.schemaLocation=/Users/gmazzola/Documents/glovo-customer-android/payments/build/roomSchemas/kspDebugKotlinTemp` we observe an absolute path, which invalidates the whole propose of the `KspWorkaround`.

I did not test with KAPT, but it seems it's also affected by the issue.

## Fix: use relative paths instead
The code was modified to provide a relative path to the project instead.

### DIFF
--- a/src/main/groovy/org/gradle/android/workarounds/room/argumentprovider/RoomSchemaLocationArgumentProvider.groovy
+++ b/src/main/groovy/org/gradle/android/workarounds/room/argumentprovider/RoomSchemaLocationArgumentProvider.groovy
@@ -2,6 +2,7 @@ package org.gradle.android.workarounds.room.argumentprovider
 
 import groovy.transform.CompileStatic
 import org.gradle.android.workarounds.RoomSchemaLocationWorkaround
+import org.gradle.api.Project
 import org.gradle.api.file.Directory
 import org.gradle.api.provider.Provider
 import org.gradle.api.tasks.Internal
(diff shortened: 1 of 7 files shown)


src/main/groovy/org/gradle/android/workarounds/room/argumentprovider/RoomSchemaLocationArgumentProvider.groovy
```diff
@@ -11,20 +12,25 @@ import org.gradle.process.CommandLineArgumentProvider
 
 @CompileStatic
 abstract class RoomSchemaLocationArgumentProvider implements CommandLineArgumentProvider {
+
+    @Internal
+    final File projectDir;
```
gmazzo, line 17: Storing the `projectDir` instead of the `Project` to be compatible with the configuration cache.

cdsap:
Hi @gmazzo, thank you for the contribution, really appreciate the effort. 

I was able to reproduce the issue in the project `nowinandroid` with Ksp [1.8.20-1.0.10](https://github.com/google/ksp/releases/tag/1.8.20-1.0.10).

We analyzed the issue because the Ksp Workaround was tested in cache-relocatable scenarios, after reviewing the code we suspect that the origin is in this change: https://github.com/google/ksp/commit/76ac8de5cd4769b2cfd7c80ad156309c5f27d52a#diff-a345032d6c965be6e81478437767cb750c83bd0a952014661e86d454ea5a342b where Ksp populates the plugin options with the existing command line argument providers set in the workaround. 

This change is present in version 1.8.10-1.0.9,  we are reviewing carefully the Ksp changes to address correctly this issue.

I tested the Kapt workaround and is not affected by this issue.

As soon we have a plan I let you know if we want to proceed with this PR. 

Thanks!

cdsap:
Hi @gmazzo, 
We didn't have feedback for the issue https://github.com/google/ksp/issues/1359 in Ksp, I created a PR https://github.com/google/ksp/pull/1368 and also waiting for feedback.
Going to close this PR. Thanks again for the great contribution

PR #485, opened 2023-04-11
Add release notes to published plugin description and read version file using CC-comaptible api

### LINKED ISSUE #437: User can read release notes in the plugin portal
We should use the `description` field of the plugin declaration to specify release notes. Currently, it is always a static string. 

The `description` field is shown on the plugin portal for each published version. Setting the `description` field to release notes means they can be seen directly in the plugin portal, rather than only on GitHub. This makes release notes and changes much more discoverable for users wanting to update.

An example of this can be seen on the plugin portal page for `com.gradle.plugin-publish`: https://plugins.gradle.org/plugin/com.gradle.plugin-publish

tylerbertrand:
Resolves #437 

### Add release notes to published plugin description

* This allows users to view release notes from the plugin portal and gives them an idea of what will be gained by updating to that version
* Leverage providers.fileContents() to support configuration caching when reading release notes file

### Use configuration-cache-compatible Provider API to read version file
* Use configuration-cache-compatible Provider.fileContents() to read version file
* Inline gitReleaseTag()
* Use githubRelease extension object to get tagName value for createReleaseTag task
* Correct githubRelease extension block indent from 3 spaces to 4

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -13,8 +13,10 @@ plugins {
 
 group = 'org.gradle.android'
 version = layout.projectDirectory.file("release/version.txt").asFile.text.trim()
+description = 'A Gradle plugin to fix Android caching problems'
 
 def isCI = (System.getenv('CI') ?: 'false').toBoolean()
+def releaseNotes = releaseNotes()
 
 repositories {
     google()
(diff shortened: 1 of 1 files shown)


etiennestuder (approved):
LGTM

build.gradle
```diff
@@ -13,8 +13,10 @@ plugins {
 
 group = 'org.gradle.android'
 version = layout.projectDirectory.file("release/version.txt").asFile.text.trim()
+description = 'Gradle plugin to fix Android caching problems'
```
etiennestuder, line 16: For consistency with our other plugins, I would say `A Gradle plugin...'.

tylerbertrand: Done

etiennestuder (approved):
LGTM

cdsap (approved):

PR #494, opened 2023-04-21
Update testing version to 8.1.0-beta01

runningcode:
Update testing version to 8.1.0-beta01

With the new version, we have noticed a test failing for the combination of AGP 8.1.0-beta01 and KGP 1.6.21. The test is using databinding and bringing dependencies from Kotlin 1.8.x.  The error is:
```
 Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 1.8.0, expected version is 1.6.0.	
```
We have created issue https://issuetracker.google.com/issues/279710208 
and we are ignoring the combination AGP 8.1.0-beta01 - KGP 1.6.21 for the failing tests


runningcode:
@cdsap Looking at the errors, it looks like when compiling against AGP 8.1, the Kotlin compile version is implicitly bumped to 1.8 which makes it incompatible.
https://ge.solutions-team.gradle.com/s/o2kr2ruajnya2/tests/:testAndroid8_1_0_beta01/org.gradle.android.RoomSchemaLocationWorkaroundTest/schemas%20are%20generated%20into%20task-specific%20directory%20and%20are%20cacheable%20with%20kotlin%20and%20kapt%20workers%20enabled%20(Android%208.1.0-beta01)%20(Kotlin%201.6.21)?focused-exception-line=0-137&top-execution=1

I'm surprised that this wasn't already happening with AGP 8.0 🤔 and Gradle 8. Do you have any thoughts there?

PR #497, opened 2023-04-24
Foojay toolchain plugin is applied to applicable test builds

clayburn:


cdsap (approved):

runningcode:
This is interesting. I would not want the tests themselves to rely on auto-provisioning behavior. These tests could then take an extremely long time to execute if the toolchain is provisioned during the execution of the test.

runningcode:
Yes, I agree with the motivation and that makes sense. Can we disable auto-provisioning for our CI builds to ensure that the CI machines are set up correctly and that we don't spend the build time downloading the correct toolchains?

PR #527, opened 2023-06-27
fix incorrect gav logic in the migration

cdsap:
During the kts migration, I wrongly changed:
```
def isProdPortal = !System.getProperty('gradle.portal.url')
```
by 
```
val isProdPortal = System.getProperty("gradle.portal.url") != null
```

This is causing a change in the group value when publishing the plugin in the release workflow. This pr applies the same logic


cdsap:
merging, I need to release a new version

PR #531, opened 2023-06-30
Use latest versions of AGP for main build

bigdaz:
It's not clear why we are using older versions of these libraries in `compileOnly` and `testImplementation` configurations. Newer versions should address some of the current security vulnerabilities.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -27,8 +27,8 @@ repositories {
 
 dependencies {
     val versions = mapOf(
-        "agp" to "8.1.0-rc01",
-        "sdkBuildTools" to "31.1.0-rc01",
+        "agp" to "8.0.2",
+        "sdkBuildTools" to "31.0.2",
         "spock" to "2.3-groovy-3.0",
     )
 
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -27,8 +27,8 @@ repositories {
 
 dependencies {
     val versions = mapOf(
-        "agp" to "7.3.1",
-        "sdkBuildTools" to "30.0.4",
+        "agp" to "8.1.0-rc01",
```
cdsap, line 30: can we use instead of the rc, the latest stable? I think is 8.0.2. 

Then, to have more context on why we are behind the latest version: 
https://gradle.slack.com/archives/C011SB27MA6/p1676326060418549 
The initial concern was not supporting older AGP versions correctly, but Clay's update to 7.3.1 showed no issues in the 7.x API.
I don't see any issue now updating to the latest one.

cdsap: updated to:
```
        "agp" to "8.0.2",
        "sdkBuildTools" to "31.0.2",
```

cdsap (approved):

build.gradle.kts
cdsap: @bigdaz merging the pr

PR #557, opened 2023-08-05
Minor tweaks

Goooler:
- Clean up build logic.
- Migrates `CreateGitTag` from Groovy to Kotlin.
- Polish flags, some of them are the default, ~~and seem we can enable `configureondemand`?~~
- Use build-tools 34.0 on CI.

### DIFF
--- a/.github/workflows/wrapper-upgrade-execution.yml
+++ b/.github/workflows/wrapper-upgrade-execution.yml
@@ -9,6 +9,7 @@ jobs:
   upgrade_wrapper:
     name: Execution
     runs-on: ubuntu-latest
+    if: github.repository == 'gradle/android-cache-fix-gradle-plugin'
     steps:
       - name: Set up Git configuration
         env:
--- a/buildSrc/src/main/kotlin/CreateGitTag.kt
+++ b/buildSrc/src/main/kotlin/CreateGitTag.kt
@@ -9,40 +9,38 @@ import org.gradle.work.DisableCachingByDefault
 import javax.inject.Inject
 
 @DisableCachingByDefault(because = "Produces no cacheable output")
-abstract class CreateGitTag extends DefaultTask {
+abstract class CreateGitTag @Inject constructor(
+    private val objects: ObjectFactory,
+    private val execOperations: ExecOperations
+) : DefaultTask() {
 
-    @Inject
-    abstract ObjectFactory getObjects()
+    @get:Input
+    val tagName: Property<String> = objects.property(String::class.java)
 
-    @Inject
-    abstract ExecOperations getExecOperations()
-
-    @Input
-    abstract Property<String> getTagName()
-
-    @Input
-    @Optional
-    final abstract Property<Boolean> overwriteExisting = objects.property(Boolean).convention(false)
+    @get:Input
+    @get:Optional
+    val overwriteExisting: Property<Boolean> = objects.property(Boolean::class.java).apply {
+        value(false)
+    }
 
     @TaskAction
-    def tag() {
+    fun applyArgbash() {
         logger.info("Tagging HEAD as ${tagName.get()}")
-        execOperations.exec { execSpec ->
-            def args = ["git", "tag"]
+        execOperations.exec {
+            val args = mutableListOf("git", "tag")
             if (overwriteExisting.get()) {
                 args.add("-f")
             }
             args.add(tagName.get())
-            execSpec.commandLine(args)
+            commandLine(args)
         }
-        execOperations.exec { execSpec ->
-            def args = ["git", "push", "origin"]
+        execOperations.exec {
+            val args = mutableListOf("git", "push", "origin")
             if (overwriteExisting.get()) {
                 args.add("-f")
             }
             args.add("--tags")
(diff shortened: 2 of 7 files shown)


runningcode:
Thanks for the contribution, could you please add a description with the motivation/benefits for this PR?

Goooler:
Done.

runningcode:
Thanks for adding the description.
[In the link](https://docs.gradle.org/current/userguide/organizing_gradle_projects.html#sec:build_sources) you added, it states `The directory buildSrc is treated as an included build.`.

This means that the migration you performed would not change anything. Are there any other reasons for doing this?

Goooler:
Just see `The directory buildSrc is treated as an included build`, seems it's landed well. Actually there are a few minor changes except 5079b0db99c58ed651ce8e9bb4f8b600cb6eacdb, can we retain them?

runningcode:
Yes, some of the other changes make sense. Could you revert the changes to the `buildSrc` and then I will review this?

runningcode (changes requested):
Please revert the changes to migrate `buildSrc`.

gradle.properties
```diff
@@ -1,8 +1,7 @@
-org.gradle.vfs.watch=true
-org.gradle.daemon=true
 org.gradle.parallel=true
 org.gradle.caching=true
 org.gradle.configuration-cache=true
+org.gradle.configureondemand=true
```
runningcode, line 4: can you tell us more about the changes to the gradle flags here and why there are being made?

runningcode (changes requested):
Thanks again for the contribution and the ideas to improve this repository. 🙏 

I think there are 3 separate things that this PR does.
1. cleans up build logic
2. migrates build logic from groovy to kotlin
3. changes gradle daemon flags.

I agree with 1, the build logic clean up
2. we need to discuss internally if this is a change we want.
3. initially does not seem to be a beneficial change.

Could you please update the PR description and also provide a description of why 2 and 3 are being done?

gradle.properties
Goooler: I see it was disabled by default, and suggest to be enabled in https://docs.gradle.org/8.2.1/userguide/build_environment.html#sec:gradle_configuration_properties

runningcode: I don't see a suggestion to enable this and this feature has no benefit in a single project build. Could you please undo this change.

Goooler: Done.

gradle.properties
```diff
@@ -1,5 +1,3 @@
-org.gradle.vfs.watch=true
-org.gradle.daemon=true
```
runningcode, line 2: why did we disable the Gradle daemon?

runningcode (changes requested):

gradle.properties
runningcode: Thanks!

gradle.properties
Goooler: Cause they are the default for now.

buildSrc/src/main/kotlin/CreateGitTag.kt
```diff
@@ -9,40 +9,39 @@ import org.gradle.work.DisableCachingByDefault
 import javax.inject.Inject
 
 @DisableCachingByDefault(because = "Produces no cacheable output")
```
runningcode, line 11: In order to reduce duplication in our codebase, can you make sure that the kotlin version of this file looks the same as this file?

https://github.com/gradle/gradle-enterprise-build-validation-scripts/blob/main/buildSrc/src/main/kotlin/CreateGitTag.kt

gradle.properties
runningcode: Thanks. This looks good.

runningcode (changes requested):
One more comment and we are good to merge this!

buildSrc/src/main/kotlin/CreateGitTag.kt
Goooler: c64ccba1917fe678ed255f888bbdbb05df2dcea4

.github/workflows/wrapper-upgrade-execution.yml
```diff
@@ -9,6 +9,7 @@ jobs:
   upgrade_wrapper:
     name: Execution
     runs-on: ubuntu-latest
+    if: github.repository == 'gradle/android-cache-fix-gradle-plugin'
```
Goooler, line 12: This job fails in forked repos, see https://github.com/Goooler/android-cache-fix-gradle-plugin/actions/runs/5816176844

runningcode: Thanks for noting this. I would either remove this workflow from your fork or setup the correct credentials for this job in your fork.
Adding this check to this repository is noise in the codebase. We can't be worrying about forks.

Goooler: And I believe this job should not be executed in forks, it will send commits like 30480245dbfbf94c5dd227002f7a48a9c9bc498a.

runningcode (approved):
Thanks for the PR!

PR #566, opened 2023-08-18
PTS improvements

ribafish:
- Enabled Predictive Test Selection (PTS) by default.
- Set the `FAST` PTS profile for `testAndroid${version}` test tasks and `STANDARD` profile for "regular" `test` task.
- Set `RELEVANT_TESTS` PTS mode for PRs (pre-merge) and `REMAINING_TESTS` for `main` (post-merge) branch builds.
- Enabled `Verify Build` action for all PRs, not just the ones targeting `main`


release/changes.md
```diff
@@ -1 +1,5 @@
-- [NEW] Run tests against AGP 8.2.0-alpha16
\ No newline at end of file
+- [NEW] Run tests against AGP 8.2.0-alpha16
+- [NEW] Enabled Predictive Test Selection (PTS) by default.
```
runningcode, line 2: I don't think any of these PTS related items should be in the release notes since it doesn't affect consumers in any way.

runningcode (changes requested):
I like where this is going, but I don't like the extra complexity that adding the PTS Mode adds to this already complex CI build. Is there a way we can do it without adding the extra dependency on the `generate_versions_pts` job?

Second thing, if we enable PTS by default, let's be sure to disable it for release jobs.

ribafish:
> I like where this is going, but I don't like the extra complexity that adding the PTS Mode adds to this already complex CI build. Is there a way we can do it without adding the extra dependency on the `generate_versions_pts` job?

I don't see a different way to be honest, at least not a simpler one. This seems pretty straightforward to me though, If this run has `github.ref_name` as `main`, then it's `REMAINING_TESTS`, otherwise it's `RELATIVE_TESTS`. The only thing to note is that when a PR is opened and triggers this action, the `ref_name` will hold the automatically created "merge" branch, so if it's a PR, this won't be true, even if base branch is `main`.

> Second thing, if we enable PTS by default, let's be sure to disable it for release jobs.

The way it is set up at the moment, the only way I see that wouldn't create a completely separate workflow for release jobs (which might also make sense to do tbh) is to use `workflow_dispatch.inputs` with default set to false and trigger it without it (by default), while still allowing to turn it on if wanted when running it manually and set it to true if it's empty (which happens when the action is started from push to main or PR)

.github/workflows/build-verification.yml
```diff
@@ -48,19 +59,24 @@ jobs:
...
+        run: >-
+            ./gradlew test
+            -x signPluginMavenPublication
+            -x signAndroidCacheFixPluginPluginMarkerMavenPublication
+            -Porg.gradle.java.installations.auto-download=false
+            -Dpts.mode=${{ needs.generate_versions_pts.outputs.pts-mode }}
+            -PisPTSEnabled=${{ github.event.inputs.pts_enabled }}
         env:
           GRADLE_ENTERPRISE_ACCESS_KEY: ${{ secrets.GE_SOLUTIONS_ACCESS_TOKEN }}
-          ORG_GRADLE_PROJECT_isPTSEnabled: ${{ github.ref_name != 'main' }}
```
runningcode, line 54: Is it possible to use variable here to set the remaining tests mode inline instead of having an action to set it?

something like `pts.mode: if ${{ github.ref_name == 'main' }}; then 'REMAINING_TESTS'; else 'RELEVANT_TESTS'` and then pass that value?

ribafish: `if` statements can't be used in the `ENV` part of actions, but there are some tertiary workarounds, one of which I've applied. It does need to be duplicated on both test jobs though...

release/changes.md
```diff
@@ -1,2 +1,2 @@
 - [NEW] Run tests against AGP 8.2.0-alpha16
-- [NEW] Run tests against AGP 8.1.1
```
runningcode, line 2: why is this showing a diff? is there a whitespace change?

.github/workflows/build-verification.yml
runningcode: Thanks for looking in to this. While there is some duplication in this solution, i prefer the simplicity of it. It also makes it easier to use as an example for customers.

.github/workflows/build-verification.yml
```diff
@@ -4,8 +4,13 @@ on:
   push:
     branches: [ main, citest ]
   pull_request:
-    branches: [ main ]
   workflow_dispatch:
+      inputs:
```
runningcode, line 8: do we still need this?

ribafish: Unless we decided to keep it on at all times, for staging/release builds as well, then yes, this is needed.

release/changes.md
ribafish: Yeah, I think it was without the newline at the end before. It's weird though, as I resolved it to the main branch variant of it when merging..

.github/workflows/build-verification.yml
runningcode: got it!

.github/workflows/build-verification.yml
```diff
@@ -4,8 +4,13 @@ on:
   push:
     branches: [ main, citest ]
   pull_request:
-    branches: [ main ]
```
runningcode, line 7: can you remind me why this change was made?

runningcode (approved):

.github/workflows/build-verification.yml
runningcode: its meant to disable pts when triggered from workflow dispatch.

runningcode (commented):
Alright as discussed and [as per the manual](https://docs.gradle.com/enterprise/predictive-test-selection/#relevant-vs-remaining-tests), Let's use remaining tests mode for all pre-release cases instead of disabling PTS for those cases.

runningcode (approved):
Looks great. I like the simplicity of this approach.

.github/workflows/build-verification.yml
```diff
@@ -16,7 +15,7 @@ jobs:
       - id: setup-matrix
         run: echo "matrix=$(jq -cM '.supportedVersions | keys' src/main/resources/versions.json | sed -e 's/\./_/g' -e 's/-/_/g')" >> $GITHUB_OUTPUT
       - name: debug
-        run: echo ${{ steps.setup-matrix.outputs.matrix }}
+        run: echo "matrix=${{ steps.setup-matrix.outputs.matrix }}"
```
runningcode, line 18: just curious, why this change?

ribafish: Oh, it's still left from the first variant where I set the PTS mode in that job, but I still left it in as it's more verbose when looking at logs. I can remove this change if wanted.

runningcode: let's leave it how it originally was to simplify and reduce the changeset.

Issue #508, opened 2023-05-11
Consider removing version check in favor of version compatibility matrix


runningcode:
We emit a warning if the Android Cache fix plugin is run against a newer version of AGP. We should consider removing the version check in favor of having a version compatibility matrix.

runningcode:
We should write the version matrix in the readme and this way we don't need to release new versions whenever a new version is released.

PR #567, opened 2023-08-19
Removing warning compatibility check

cdsap:
Issue #508 
Remove the compatibility check generating the warnings in the Plugin logic. Adding compatibility supported AGP versions in  README.


clayburn (approved):

README.md
```diff
@@ -17,13 +17,26 @@
...
 * Supported Gradle versions: 7.0+
-* Supported Android Gradle Plugin versions: 7.0.4, 7.1.3, 7.2.2, 7.3.1, 7.4.2, 8.0.2, 8.1.0 and 8.2.0-alpha16
+* Supported Android Gradle Plugin versions: 7.0+
 * Supported Kotlin versions: 1.7.0+
 
-We only test against the latest patch versions of each minor version of Android Gradle Plugin.  This means that although it may work perfectly well with an older patch version (say 7.0.1), we do not test against these older patch versions, so the latest patch version is the only version from that minor release that we technically support.
-
 The Android cache fix plugin is compatible with the [Gradle Configuration Cache](https://docs.gradle.org/current/userguide/configuration_cache.html#header) when used in conjunction with Android Gradle Plugin 4.2.0 and above.  Using the configuration cache with earlier versions of the Android Gradle Plugin is not supported.
 
+### Compatibility matrix by AGP version
```
erichaagdev, line 25: I find this matrix and section confusing. Above we already say `Supported Android Gradle Plugin versions: 7.0+` so do we really need this?

README.md
```diff
@@ -131,6 +144,23 @@ static class AndroidJavaCompile_BootClasspath_Workaround implements Workaround {
 }
 ```
 
+### Compatibility matrix by AGP version and Workaround
```
erichaagdev, line 147: Why aren't these workarounds supported for newer versions of AGP? Or is that they are fixed and so the workarounds are not needed?

src/main/groovy/org/gradle/android/AndroidCacheFixPlugin.groovy
```diff
@@ -55,14 +55,10 @@ class AndroidCacheFixPlugin implements Plugin<Project> {
...
-        workarounds.addAll(initializeWorkarounds(project))
+        workarounds.addAll(initializeWorkarounds())
 
-        if (!isSupportedAndroidVersion(project)) {
-            if (isMaybeSupportedAndroidVersion(project)) {
-                Warnings.MAYBE_SUPPORTED_ANDROID_VERSION.warnOnce(project)
-            } else {
-                throw new RuntimeException("Android plugin ${CURRENT_ANDROID_VERSION} is not supported by Android cache fix plugin. Supported Android plugin versions: ${SUPPORTED_ANDROID_VERSIONS.join(", ")}. Please check if a newer version of this plugin is available or override with -D${IGNORE_VERSION_CHECK_PROPERTY}=true.")
-            }
+        if (!isMaybeSupportedAndroidVersion()) {
```
erichaagdev, line 60: What does it mean to be "maybe supported"? Something should be either supported or not. Is there a scenario where this returns `true` and we later find out the version is not supported?

erichaagdev (changes requested):

src/main/groovy/org/gradle/android/AndroidCacheFixPlugin.groovy
cdsap: This is not something introduced by these changes/pr. 
The explanation "maybe supported" comes from the idea that release cycles of the AGP won't include api changes in patch releases. Our mechanism to apply workarounds in the project build classpath is based on compatibility with major-minor AGP versions, hence the previous approach to throw a warning when the version didn't match with major-minor-patch.  
If it's not clear, we can rename the function to isSupportedAndroidVersion

cdsap: exactly, some of the workarounds are fixed, each workaround contains the annotation `AndroidIssue` where you can track when it was fixed in the AGP. Example:
https://github.com/gradle/android-cache-fix-gradle-plugin/blob/main/src/main/groovy/org/gradle/android/workarounds/BundleLibraryClassesWorkaround.groovy#L13C2-L13C14

README.md
cdsap: exactly, some of the workarounds are fixed, each workaround contains the annotation `AndroidIssue` where you can track when it was fixed in the AGP. Example:
https://github.com/gradle/android-cache-fix-gradle-plugin/blob/main/src/main/groovy/org/gradle/android/workarounds/BundleLibraryClassesWorkaround.groovy#L13C2-L13C14

README.md
cdsap: I found that adding the compatibility table in the first section adds a bit of verbosity to the initial plugin description. 
The initial description contains general build stack tooling information and BT features compatibilities(conf_cache). 

The new subsection contains specific information about our release cycle related to the AGP versions tested and supported by the plugin. 

Don't have a strong preference here and open for suggestions/ideas

runningcode: This does seem quite redundant. The words "compatible" and "supported" are quite similar and it isn't clear where we are using what.

README.md
```diff
@@ -17,13 +17,26 @@
 Some Android plugin versions have issues with Gradle's build cache feature. When applied to an Android project this plugin applies workarounds for these issues based on the Android plugin and Gradle versions. For other versions, please see [older versions.](#older-android-gradle-plugin-versions)
 
 * Supported Gradle versions: 7.0+
-* Supported Android Gradle Plugin versions: 7.0.4, 7.1.3, 7.2.2, 7.3.1, 7.4.2, 8.0.2, 8.1.0 and 8.2.0-alpha16
+* Supported Android Gradle Plugin versions: 7.0+
```
runningcode, line 20: Did you write in the doc that there is also a max supported version? I added a comment in the doc, but what are the pros and cons of having that max version?
If we do have a max version, we need to document it.

README.md
runningcode: We used to point users to each workaround's source for documentation on what workaround is applied to what version. I'm kinda neutral on whether nor not this is a good idea since it does add maintentance cost.

This chart isn't as precise as the code since the workarounds applied in the code are applied to specific major/minor/patch versions and are also fixed in major/minor/patch versions.

This is a chart of workarounds applied to AGP versions.
```suggestion
### AGP version and Workarounds applied
```

src/main/groovy/org/gradle/android/AndroidCacheFixPlugin.groovy
runningcode: The maybe confused me too. I think we should drop it.

src/main/groovy/org/gradle/android/AndroidCacheFixPlugin.groovy
```diff
@@ -55,14 +55,10 @@ class AndroidCacheFixPlugin implements Plugin<Project> {
...
+            throw new RuntimeException("Android plugin ${CURRENT_ANDROID_VERSION} is not supported by Android cache fix plugin. Supported Android plugin versions: ${SUPPORTED_ANDROID_VERSIONS.join(", ")}")
```
runningcode, line 61: Once we resolve the discussion on what max/min ranges are for this, I think we should update this error message to be more precise.

runningcode (commented):
Added some feedback. After I published, I realized this PR is in draft mode. Let's resolve the discussion in the spec first before going to the PR to avoid back and forth on the PR.

src/main/groovy/org/gradle/android/AndroidCacheFixPlugin.groovy
cdsap: done

README.md
cdsap: compatibility matrix removed

README.md
cdsap: after discussion in the doc, we dropped the max supported version and the patch labels on the supported android versions

src/main/groovy/org/gradle/android/AndroidCacheFixPlugin.groovy
cdsap: discussion on max/min resolved, what are your suggestions for this message?

runningcode: In this case this means that the AGP version is too old. In that case I think we should recommend that they use the Android Cache Fix plugin which is compatible with their version. What do you think about: `Android plugin ${CURRENT_ANDROID_VERSION} is not supported by Android cache fix plugin. For older Android Gradle Plugin versions, please use Android Cache Fix Plugin 2.4.6`

src/main/groovy/org/gradle/android/Versions.groovy
```diff
@@ -43,33 +40,13 @@ class Versions {
...
-
-    static boolean isSameMajorAndMinorAsSupportedVersion() {
-        return isSameMajorAndMinorAsSupportedVersion(CURRENT_ANDROID_VERSION)
-    }
-
-    static boolean isSameMajorAndMinorAsSupportedVersion(VersionNumber versionNumber) {
-        return SUPPORTED_ANDROID_VERSIONS.stream().anyMatch { supportedVersion ->
-            versionNumber.major == supportedVersion.major && versionNumber.minor == supportedVersion.minor
-        }
+    static boolean isSupportedAndroidVersion() {
```
runningcode, line 49: This is a nice simplification

src/main/groovy/org/gradle/android/AndroidCacheFixPlugin.groovy
cdsap: done

PR #587, opened 2023-09-25
Update agp 8.2.0-beta05

cdsap:
https://androidstudio.googleblog.com/2023/09/android-studio-hedgehog-beta-5-now.html


src/main/resources/versions.json
```diff
@@ -3,7 +3,7 @@
         "8.3.0-alpha03": [
```
runningcode, line 3: isn't this also at alpha05 now?

runningcode (commented):
great that there are now less places to update!

runningcode (approved):

runningcode (approved):

src/main/resources/versions.json
cdsap: you are right, alpha05 is available, pushed

PR #588, opened 2023-09-28
Update tests to execute against AGP 8.1.2 and AGP 8.2.0-beta06

runningcode:


runningcode:
readme needs update as well

runningcode:
scratch that. readme doesn't have specific versions in it anymore.

PR #593, opened 2023-10-05
Update supported versions to 8.4

cdsap:
Supported version update to Gradle 8.4


src/main/resources/versions.json
```diff
@@ -1,7 +1,7 @@
 {
     "supportedVersions": {
         "8.3.0-alpha05": [
-            "8.3"
+            "8.4"
         ],
         "8.2.0-beta06": [
             "8.2.1"
```
runningcode, line 7: I think we should also test this beta and as well as the latest stable version 8.1.2 against Gradle 8.4.

cdsap: sounds good, [01b38f9](https://github.com/gradle/android-cache-fix-gradle-plugin/pull/593/commits/01b38f90814d75b182e31bc6a2db0396012df833)

PR #650, opened 2024-01-29
Updated to AGP 8.4.0-alpha07

ribafish:
Needed to update Gradle to `8.6-rc01`, even though I didn't find it in release notes because of:

```
Minimum supported Gradle version is 8.6-rc-1. Current version is 8.5.
```


runningcode (approved):

src/main/resources/versions.json
```diff
@@ -1,7 +1,7 @@
 {
     "supportedVersions": {
-        "8.4.0-alpha02": [
-            "8.5"
+        "8.4.0-alpha07": [
+            "8.6-rc-1"
```
runningcode, line 4:
```suggestion
            "8.6-rc-3"
```
Why use rc 1 if rc 3 is out?

runningcode (changes requested):

src/main/resources/versions.json
ribafish: Autopilot 🤦

PR #676, opened 2024-03-28
Use provider API so that version.json is configuration cache input

runningcode:
Previously when making changes to the version.json file would not cause the configuration cache to re-execute since the contents of the versions.json file were not inputs to configuration cache. As an example, if you bumped the version from  8.4.0-beta01 to 8.4.0-beta02, the task name would still be `beta01` since the configuration cache would re-run.

This also bumps the supported version to 8.4.0-beta02

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -234,6 +234,7 @@ fun releaseNotes(): Provider<String> {
 
 @Suppress("UNCHECKED_CAST")
 fun getSupportedVersions(): Map<String, Array<String>> {
+    val versionFile = providers.fileContents(layout.projectDirectory.file("src/main/resources/versions.json"))
     return (JsonSlurper()
-        .parse(providers.fileContents(layout.projectDirectory.file("src/main/resources/versions.json")).asBytes.get()) as Map<String, Map<String, Array<String>>>).getValue("supportedVersions")
+        .parse(versionFile.asBytes.get()) as Map<String, Map<String, Array<String>>>).getValue("supportedVersions")
 }


build.gradle.kts
```diff
@@ -235,5 +235,5 @@ fun releaseNotes(): Provider<String> {
 @Suppress("UNCHECKED_CAST")
 fun getSupportedVersions(): Map<String, Array<String>> {
     return (JsonSlurper()
-        .parse(file("src/main/resources/versions.json")) as Map<String, Map<String, Array<String>>>).getValue("supportedVersions")
+        .parse(providers.fileContents(layout.projectDirectory.file("src/main/resources/versions.json")).asBytes.get()) as Map<String, Map<String, Array<String>>>).getValue("supportedVersions")
```
erichaagdev, line 238: "Applying suggestions on deleted lines is not supported."

Maybe this is more readable?

```kotlin
@Suppress("UNCHECKED_CAST")
fun getSupportedVersions(): Map<String, Array<String>> {
    val versions = providers.fileContents(layout.projectDirectory.file("src/main/resources/versions.json"))
    return (JsonSlurper().parse(versions.asBytes.get()) as Map<String, Map<String, Array<String>>>).getValue("supportedVersions")
}
```

erichaagdev: I tried to figure out a way to remove the need to hardcode `src/main/resources` You can, but it's quite verbose. You can call `sourceSets.main.map { it.resources .... }`.

runningcode: I like the first suggestion and I applied it.
I also had issues with github suggestions in PR comments today. It wasn't working.

PR #682, opened 2024-04-05
Test against 8.4.0-rc02

runningcode:


ribafish (changes requested):
Shouldn't we also update (which I get someone forgot to do for beta02 as well)? https://github.com/gradle/android-cache-fix-gradle-plugin/blob/834ccdfe192ba30abaa2a68dfc786b4f0db63357/src/test/groovy/org/gradle/android/WorkaroundTest.groovy#L16

runningcode:
That was me forgetting on the last update. Thanks for catching that :)

src/test/groovy/org/gradle/android/WorkaroundTest.groovy
```diff
@@ -12,8 +12,8 @@ class WorkaroundTest extends Specification {
         workarounds.collect { it.class.simpleName.replaceAll(/Workaround/, "") }.sort() == expectedWorkarounds.sort()
         where:
         androidVersion  | expectedWorkarounds
-        "8.5.0-alpha01" | ['JdkImage']
-        "8.4.0-alpha13" | ['JdkImage']
+        "8.5.0-alpha02" | ['JdkImage']
```
ribafish, line 15:
```suggestion
        "8.5.0-alpha03" | ['JdkImage']
```

ribafish (changes requested):

ribafish (approved):

PR #717, opened 2024-06-07
Update Link to Room Gradle Plugin Release

clayburn:
Since 2.6.0-alpha02, the Room Gradle Plugin is available. But now that Room 2.6.0 (and later) are released, it would be more appropriate to point users to the 2.6.0 major release containing this plugin.

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,7 +145,7 @@ Not fixed by the Android Cache Fix plugin since it has no workaround but is fixe
 
 ### Room
 The Room annotation processor causes cache misses: https://issuetracker.google.com/issues/132245929.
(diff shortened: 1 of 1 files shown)


clayburn:
Actually, disregard this. I see that the release notes for 2.6.0-alpha02 provide better docs on how to use the plugin than the 2.6.0 notes do.

runningcode:
Thanks for looking in to this. Weird that they don't have a better explanation of how to apply it

erichaagdev:
What's wrong with: https://developer.android.com/jetpack/androidx/releases/room#gradle-plugin

runningcode:
Oh, yes, that's exactly what we need.

PR #750, opened 2024-09-04
NOTICE file is added and included in all assembled JAR files along with LICENSE

erichaagdev:
This PR adds a NOTICE file to the root of the project. The NOTICE file includes the licenses of all runtime dependencies.

Both the NOTICE and LICENSE files are then included in all assembled JAR files.


build.gradle.kts
```diff
@@ -113,6 +113,13 @@ publishing {
     }
 }
 
+tasks.withType<Jar> {
```
etiennestuder, line 116: Same question re: `configureEach` as on the the CCUD PR.

etiennestuder (changes requested):

build.gradle.kts
erichaagdev: Yes. It should be `configureEach`. Fixed in most recent changes. b6374fd77df8f9e7d10b6f2dade6727ed2f3055b

etiennestuder (approved):

PR #768, opened 2024-10-28
Add additional tags to the published pom

welandaz:


build.gradle.kts
```diff
@@ -104,6 +104,28 @@ publishing {
     publications {
         create<MavenPublication>("pluginMaven") {
             groupId = pluginGroupId
+            pom {
+                name.set("Android Cache Fix Gradle Plugin")
+                description.set("A Gradle Plugin to workaround cache issues in Android builds")
```
erichaagdev, line 109: I would lowercase `plugin` in the description. I think it's fine capitalized in the name because even though we typically will stylize it as `Android Cache Fix Gradle plugin` [it _is_ capitalized in titles](https://github.com/gradle/android-cache-fix-gradle-plugin?tab=readme-ov-file#android-cache-fix-gradle-plugin).
```suggestion
                description.set("A Gradle plugin to workaround cache issues in Android builds")
```

erichaagdev (approved):

PR #1741, opened 2025-01-02
Update JdkImageWorkaround.groovy to reflect that the issue was fixed in AGP 7.4.0-alpha07

kmayoral:
A minor update updating the @AndroidIssue annotation to clarify that the issue is no longer present in newer versions of AGP.


cdsap:
Hi @kmayoral. 
While the original issue indicates it was fixed in AGP 7.4.0-alpha07, our records show that we can still reproduce this issue with AGP 8.x versions, as shown in this image from last year's investigation:

Screenshot 2025-01-02 at 2 13 20 PM

(red cells represent cache misses on the java version matrix)

That's why we are not considering this issue fully resolved.

kmayoral:
Thanks so much for the quick response @cdsap ! I'll close this PR out now in that case. Thanks!

PR #1742, opened 2025-01-03
Configure Renovate to suggest AGP version bumps

gabrielfeo:
Fully automate the bumping of AGP versions to the latest patch. While there's an [existing workflow](https://github.com/gradle/dv-solutions/actions/workflows/upgrade-android-gradle-plugin-versions.yml) for bumping the AGP versions in this project, it must be manually triggered specifying the new version. Renovate can fill this gap by suggesting updates whenever new versions come out. Other kinds of bumps are still handled by Dependabot.

### Renovate

Renovate is a tool like Dependabot but much more flexible. It has built-in ["managers"][1] such as `maven`, `gradle`, and `pip_requirements` that look for certain patterns in certain files, then query ["datasources"][2] for newer versions. No built-in manager supports the custom `versions.json` file, but Renovate allows for creating a custom regex-based manager.

### Renovate custom manager

A Renovate config file is added with a custom ["regex" manager][3]. This can instruct Renovate to parse any version on any file. The goal is to detect and update AGP versions annotated with a comment:

```
# renovate: AGP version
8.6.0
```

The "manager" config declares

- what files it should process
- what it should consider a version definition within those files
- what "datasource" it should query for newer versions
- what "versioning" (scheme) the versions adhere to
- the dependency name
- the Maven registry URL the dependency comes from

```json5
{
  // Matches AGP versions annotated with a "renovate: AGP version" comment
  customType: "regex",
  fileMatch: [
    "src/main/resources/versions\\.json5",
    "gradle\\.properties",
  ],
  matchStrings: [
    // For JSON: the first double-quoted string below the comment line, e.g. "1.0.0"
    "\/\/ renovate: AGP version\\s+?\"(?<currentValue>\\S+?)\"",
    // For properties: the value of the first property below the comment line, e.g. anyProperty=1.0.0
    "# renovate: AGP version\\s+?\\S+?=(?<currentValue>\\S+?)(?:\\s|$)",
  ],
  "datasourceTemplate": "maven",
  "versioningTemplate": "maven",
  "depNameTemplate": "com.android.tools.build:gradle",
  "registryUrlTemplate": "https://dl.google.com/dl/android/maven2/",
},
```

An example PR is available in a fork: gabrielfeo/android-cache-fix-gradle-plugin#10.

image

The last four elements are commonly set on each comment and captured with regex ([example][4]), but I opted to centralize them in the Renovate config to minimize duplication, simplify the comments and the regex that parses them.

### Changes

- Annotate AGP versions in `versions.json` with a "renovate: AGP version" comment
  - Rename the file to `versions.json5` to keep syntax highlighting support when adding comments
  - Change all `JsonSlurper`s to `LAX` mode, which is lenient about comments (Groovy has no support for JSON5)
- Add a Renovate config to bump versions in `versions.json5` and `gradle.properties` too (for the build-time check described below)
- Change plugin tests to no longer require changes on every patch bump PR. ["expectedOutcomes" JSON files][5] are now minor-specific and `WorkaroundTest` no longer specifies patch numbers. The existing workflow updated the tests automatically, but it can't be as easily done with Renovate. Thanks to @erichaagdev for this suggestion.
- Declare the latest AGP version in `gradle.properties` and fail the build if it's missing from `versions.json5`. The goal is to have Renovate create a "nudge" PR if a new major or minor comes out (`versions.json5` has major/minor bumps disabled in order to keep older minor versions in the list of "supportedVersions"). The property isn't used for anything else. Thanks to @erichaagdev for this suggestion.

These changes were tested running the Renovate GitHub App against a fork. After adding the `dv.json5` preset to the config, which is located in a private Gradle repository, they were tested with dry-runs of our self-hosted Renovate runner, which has access to that repository.

- [PRs from GitHub App on fork][6]
- [dry-run of self-hosted app][7]

[1]: https://docs.renovatebot.com/modules/manager/
[2]: https://docs.renovatebot.com/modules/datasource/
[3]: https://docs.renovatebot.com/modules/manager/regex
[4]: https://docs.renovatebot.com/modules/manager/regex/#regular-expression-capture-groups
[5]: https://github.com/gradle/android-cache-fix-gradle-plugin/tree/a7aa73ba00f87bb2cdd08e58ab8190c34b70515c/src/test/resources/expectedOutcomes
[6]: https://github.com/gabrielfeo/android-cache-fix-gradle-plugin/pulls?q=is%3Apr+author%3Arenovate%5Bbot%5D+
[7]: https://github.com/gradle/renovate-agent/actions/runs/12675352428/job/35327825387


cdsap (approved):
awesome work @gabrielfeo

clayburn (approved):

PR #1753, opened 2025-01-14
Update Renovate config to avoid JSON5

gabrielfeo:
In #1742, Renovate was set up with a more usual standard `regex` matcher configuration, which is to search for versions in lines annotated with a comment `// renovate: ...`. However, the `versions.json` file is parsed by several tools which don't support JSON5 or any form of comments, namely Groovy's `json` package and `jq`. Groovy `json` required using a lenient mode (#1742), which may suppress useful validations, while `jq` required stripping the comments beforehand (#1752). Moving away from JSON5 would be a simpler solution.

Revert the `versions.json` back from JSON5 to plain JSON and change the Renovate config to match the version properties based on the fact that their values are lists, e.g. `"8.8.0": [`. Matching the version in `gradle.properties` is also simplified to no longer require a comment.

Tested with a [dry-run of gradle/renovate-agent][2] against this repo and the Renovate GitHub app against a fork, which created an [example PR][1].

[1]: https://github.com/gabrielfeo/android-cache-fix-gradle-plugin/pull/14#issue-2787066376
[2]: https://github.com/gradle/renovate-agent/actions/runs/12771500371/job/35598810573

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -12,22 +12,16 @@
     {
       // Matches AGP versions in versions.json
       customType: "regex",
-      fileMatch: ["src/main/resources/versions\\.json"],
-      "matchStringsStrategy": "recursive",
-      "matchStrings": [
-        // Matches versions followed by a list, e.g. "8.9.0": [. See https://docs.renovatebot.com/configuration-options/#recursive
-        "\\s+?\"(?<currentValue>\\S+?)\": \\[",
+      fileMatch: [
+        "src/main/resources/versions\\.json",
+        "gradle\\.properties",
       ],
-      "datasourceTemplate": "maven",
-      "versioningTemplate": "maven",
-      "depNameTemplate": "com.android.tools.build:gradle",
-      "registryUrlTemplate": "https://dl.google.com/dl/android/maven2/",
-    },
-    {
-      // Matches AGP versions annotated with a "renovate: AGP version" comment
-      customType: "regex",
-      fileMatch: ["gradle\\.properties"],
-      matchStrings: ["# renovate: AGP version\\s+?\\S+?=(?<currentValue>\\S+?)(?:\\s|$)"],
+      "matchStrings": [
+        // In versions.json, matches keys with a list value, e.g. "8.9.0": [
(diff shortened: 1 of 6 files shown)


cdsap (approved):

.github/renovate.json5
```diff
@@ -10,33 +10,39 @@
...
+        // Matches versions followed by a list, e.g. "8.9.0": [. See https://docs.renovatebot.com/configuration-options/#recursive
+        "\\s+?\"(?<currentValue>\\S+?)\": \\[",
       ],
       "datasourceTemplate": "maven",
       "versioningTemplate": "maven",
       "depNameTemplate": "com.android.tools.build:gradle",
       "registryUrlTemplate": "https://dl.google.com/dl/android/maven2/",
     },
+    {
+      // Matches AGP versions annotated with a "renovate: AGP version" comment
```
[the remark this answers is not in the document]
gabrielfeo: Good point, yes. I've updated the config removing the need for a comment. I've also merged both managers to one and removed the "recursive" mode which ended up not being necessary for the final diff.

Tested with a [dry-run][1]

[1]: https://github.com/gradle/renovate-agent/actions/runs/12771500371/job/35598810573#step:3:944

.github/renovate.json5
```diff
@@ -10,17 +10,17 @@
   "enabledManagers": ["custom.regex"],
   "customManagers": [
     {
-      // Matches AGP versions annotated with a "renovate: AGP version" comment
+      // Matches AGP versions in versions.json
```
erichaagdev, line 13:
```suggestion
      // Matches AGP versions in versions.json and gradle.properties
```

erichaagdev (approved):
Looks good aside from this one minor thing.

.github/renovate.json5
gabrielfeo: Good catch. Better yet, I'll just remove filenames from the comment. Might've made sense when there were 2 managers, but now it's just duplicating the `fileMatch` list.

```suggestion
      // Matches AGP versions in files with custom patterns
```

PR #1779, opened 2025-02-28
Clarify distinction between "supported" and "tested" AGP versions

gabrielfeo:
`versions.json` values are currently referred to as "supported" versions throughout project source code. Even though it _is_ used to determine the minimum supported AGP version, it's not a comprehensive list of versions that the plugin supports, which may lead to confusion. For instance, the plugin supports a variety of patch versions beyond those. Mainly, the `versions.json` file determines which AGP and Gradle versions the plugin will be _**tested**_ against, not the versions it supports. 

- Change references to AGP versions throughout project source from "supported" to "tested" versions, which more accurately describes the usage of that file
- Move `versions.json` to test resources. It was only needed in the main source set because the minimum AGP version was based on its content. As the minimum is changed very infrequently, it can become hardcoded, allowing the JSON and surrounding code to be moved into the test source set.
- Update Renovate PR titles and commit subjects to a clearer "Update _tested_ AGP versions"

One property keeps "supported" in the name: the `earliestSupportedAndroidVersion` is the minimum AGP version the plugin is _tested_ against. While the minimum _supported_ version is now hardcoded, there's a test that asserts it corresponds to the minimum tested version. No changes are made to the project docs.


erichaagdev (commented):
Do we use `src/main/resources/testedVersions.json` at runtime? Can we move it to `src/test/resources`?

If so, I would just call it `versions.json` (like it was before) and change the top level object name from `testedVersions` to `versions`. 

All other renames from `supported` to `tested` lgtm

cdsap (approved):

gabrielfeo:
> Do we use `src/main/resources/testedVersions.json` at runtime? Can we move it to `src/test/resources`?
> 
> If so, I would just call it `versions.json` (like it was before) and change the top level object name from `testedVersions` to `versions`.
> 
> All other renames from `supported` to `tested` lgtm

Hi, @erichaagdev! [We do use it at runtime](https://github.com/gradle/android-cache-fix-gradle-plugin/pull/1779/files#diff-d61d4ec178ff6dab40754d86f3328b133016a3b64a23dfc312a40fbdd9265687R20), so could not move it to `src/test/resources`.

gabrielfeo:
@erichaagdev Great suggestion to hardcode the min version. The JSON is now in the test source set. Can you please take another look?

I opted to keep [this test](https://github.com/gradle/android-cache-fix-gradle-plugin/pull/1779/files#diff-afb69726636b82e3b293b7f6ae61e3d33799952dd17eb24370b89e7741451294R6) to make sure we keep testing against the minimum version (or its last stable patch). Let me know if you disagree.

erichaagdev (approved):
Great change! Thanks for doing this. 🙌

I would leave the test since it verifies the minimum tested version is the same as the actual minimum supported version. In other words, it would break if someone updated the versions file but not the hard coded value.

PR #1817, opened 2025-05-20
Updated GithubActions to use JDK 17

ribafish:
Updated GithubActions to use JDK 17. This is a required change, required by the [org.gradle.toolchains.foojay-resolver-convention](https://plugins.gradle.org/plugin/org.gradle.toolchains.foojay-resolver-convention) plugin v1.0.0. 

Without this, we either can't upgrade to the latest version of [org.gradle.toolchains.foojay-resolver-convention](https://plugins.gradle.org/plugin/org.gradle.toolchains.foojay-resolver-convention), or we have failing builds as seen on the example resolver upgrade PR: https://github.com/gradle/android-cache-fix-gradle-plugin/pull/1816

### DIFF
--- a/.github/workflows/build-verification.yml
+++ b/.github/workflows/build-verification.yml
@@ -32,7 +32,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-java@v4
         with:
-          java-version: 11
+          java-version: 17
           distribution: 'temurin'
       - name: Set up Gradle
         uses: gradle/actions/setup-gradle@v4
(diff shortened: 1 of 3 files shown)


tylerbertrand (approved):

tylerbertrand:
Any reason not to jump to 21?

ribafish:
> Any reason not to jump to 21?

No particular reason tbh. Just wanted to do a minimal update, but I'm happy to bump to 21 as well. Let's discuss in our sync.

PR #1873, opened 2025-08-22
AGP 9 update

cdsap:
## AGP 9 update
We are not dropping AGP versions in this PR. We will first discuss this with the team.

* The new version requires Gradle 9, which in turn requires JDK 17.
* Updated Java versions in the GitHub Action to set 17 as the base version.
* Added a unit test case for the generated projects.
* Removed SourceCompatibility from the builder, since the currently supported AGP versions always target JDK > 9.
* Added AGP 9 outcome files.


.github/workflows/build-verification.yml
```diff
@@ -77,15 +77,15 @@ jobs:
...
+          java-version: 17.0.16+8
           distribution: 'zulu'
       - name: Set ZULU_JDK environment variable
         run: |
           echo "ZULU_JDK=$JAVA_HOME" >> $GITHUB_ENV
       - name: Setup ZULU_ALT_JDK
         uses: actions/setup-java@v5
         with:
-          java-version: 11.0.13
+          java-version: 17.0.10+7
```
ribafish, line 88: Not an issue, just interested if there's a special reason why we're setting a specific patch, instead of simply `17`?

ribafish (approved):

.github/workflows/build-verification.yml
cdsap: We have a test that verifies the behavior of the JdkImageWorkaround for different (minor) versions of the same vendor:
https://github.com/gradle/android-cache-fix-gradle-plugin/blob/main/src/test/groovy/org/gradle/android/JdkImageWorkaroundTest.groovy#L77

Issue #447, opened 2023-02-18
Update AGP latest version 


cdsap:
Once we updated the AGP version in the classpath to `7.3.1`, we still need to point to the latest stable version of the AGP. 
https://github.com/gradle/android-cache-fix-gradle-plugin/blob/main/build.gradle#L26

The new variant API is available in recent versions and it's a pre-requisite to start reworking the workaround variant logic with types provided in the new API. 
The old variant is going to be removed in AGP 9.

runningcode:
@cdsap is there anything else we need to do here?

runningcode:
Can you check if we still have usages of the Variant API?

cdsap:
Hi, we have still the `JdkImageWorkaround` with references like:
```
            android.unitTestVariants.all(configureVariant)
            android.applicationVariants.all(configureVariant)
```

PR #1886, opened 2025-09-04
Avoid using old Variant API extending JavaCompile task

cdsap:
Since AGP 9.0.0-alpha04, the new DSL has replaced the old DSL implementation with the Variant API.
Our dependency on the old API was discussed in [gradle/android-cache-fix-gradle-plugin#447](https://github.com/gradle/android-cache-fix-gradle-plugin/issues/447).

Technically, we could use the AndroidComponent variant selector, but the workaround requires access to the JavaCompile task provider. This is a breaking change, as there is no equivalent in the new API.

AGP currently provides an opt-out flag to continue using the old behavior, which we enabled in the PR updating to Alpha04. However, this option will be removed permanently in AGP 10, so we need a long-term solution.

As an initial step, this PR configures the JDK transform directly when configuring the JavaCompile task. Further analysis and testing will be required to verify the effectiveness of this approach.


ribafish:
What is the downside of applying the workaround to all `JavaCompile` tasks (at least in projects that have the AGP applied)?

PR #1889, opened 2025-09-16
Adding Normalization to `java/lang/invoke`

cdsap:
Add runtime classpath normalization for `java/lang/invoke/**` in `JdkImageWorkaround`.

We’ve observed persistent differences on JavaCompile inputs in Android projects even after applying the Android Cache Fix Plugin. These differences stem from JDK image layout/path variations that surface under java/lang/invoke. See the prior discussion in issue #341 for background and examples.

My impression is that this approach might be a bit aggressive. Should we instead do the inverse, where the default behavior is to not apply the normalization?


clayburn:
> My impression is that this approach might be a bit aggressive. Should we instead do the inverse, where the default behavior is to not apply the normalization?

If we decide that this is too aggressive to do by default, then my opinion would be that the normalization is so simple that there is not much value in having it in the plugin behind an opt-in flag. In that case, I would just recommend that we document it in the plugin README or similar.

src/test/groovy/org/gradle/android/JdkImageWorkaroundTest.groovy
```diff
@@ -308,4 +308,103 @@ class JdkImageWorkaroundTest extends AbstractTest {
         where:
         androidVersion << TestVersions.latestAndroidVersions
     }
+
+    def "Invoke normalization property is enabled and normalizes **/java/lang/invoke/**"() {
```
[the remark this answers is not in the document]
cdsap: good point, adding your suggestion

src/test/groovy/org/gradle/android/JdkImageWorkaroundTest.groovy
```diff
@@ -308,4 +308,100 @@ class JdkImageWorkaroundTest extends AbstractTest {
...
+            ).build()
+
+        then:
+        buildResult.task(':app:testDebugUnitTest').outcome == TaskOutcome.SUCCESS
+    }
+
+    def createLibJavaClass(String label) {
+        file("library/src/main/java/com/foo/java/lang/invoke/Bar.java")
+            .createParentDirectories()
+            .newWriter { w ->
```
gabrielfeo, line 396: This is what I meant to suggest

```suggestion
            .withWriter { w ->
```

cdsap: I've applied the suggestion manually

PR #1988, opened 2026-04-07
Add AGP version upgrade instructions to repo

clayburn:
## Summary

Moves AGP version upgrade instructions from internal documentation into the repository itself, so they live alongside the code they describe and stay up to date as the process evolves.

## Alternative locations considered

This PR places the instructions at `src/test/resources/CLAUDE.md`, which makes them available as context for AI-assisted development (Claude Code picks up `CLAUDE.md` files automatically). Some alternatives worth discussing:

- **Root `CLAUDE.md`** - More discoverable for Claude Code since it's loaded at the project root, but mixes AI-specific context with project-level files.
- **Root `README.md`** - Most discoverable for human contributors, but this is process documentation rather than project overview content.

Open to moving it wherever the team prefers. The main goal is getting these instructions out of internal docs and into the repo.


gabrielfeo:
I think this is a good change in principle, but that we should

- use an `AGENTS.md` file, which isn't coupled to Claude (it can also be in the child directory), see https://agents.md/
- move these instructions to a separate MD file, e.g. `agp-version-update-instructions.md`, and briefly explain in `AGENTS.md` the cases in which the agent should refer to those instructions (i.e. AGP version major or minor updates, or any update failure). The benefit is that the `AGENTS.md` file is kept minimal and high-level, which generally improves LLM performance as it preserves the context window.

docs/agp-version-update-instructions.md
erichaagdev, line 1: This doesn't feel like the right location for this file. It's not a test resource.

clayburn: Agree, that would have a very unintended side effect

cdsap (approved):

gabrielfeo (approved):

PR #1996, opened 2026-04-14
Ungroup AGP version updates

clayburn:
## Summary
This is to help unblock the 9.1.0 → 9.1.1 update from #1989.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -53,19 +53,6 @@
       "ignoreUnstable": false,
     },
     {
-      // Customize commit message for AGP version changes
-      "matchFileNames": [
-        "src/test/resources/versions\\.json",
-        "gradle\\.properties",
-      ],
-      "commitMessageTopic": "tested AGP versions",
-    },
-    {
-      // Ignore broken 9.2.0-rc01
-      "matchDepNames": ["com.android.tools.build:gradle"],
-      "allowedVersions": "!/^9\\.2\\.0-rc01$/",
-    },
-    {
       // AGP is managed by the custom regex manager for the test matrix
       "matchManagers": ["gradle"],
       "matchDepNames": ["com.android.tools.build:gradle"],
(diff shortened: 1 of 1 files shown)


.github/renovate.json5
```diff
@@ -53,13 +53,17 @@
...
     {
-      // Group changes about AGP versions
+      // Customize commit message for AGP version changes
       "matchFileNames": [
         "src/test/resources/versions\\.json",
         "gradle\\.properties",
       ],
       "commitMessageTopic": "tested AGP versions",
-      "groupName": "tested AGP versions",
+    },
```
gabrielfeo: Removing the group name but keeping the commit message modification would result in different PRs for each version having the same title, e.g.

- "Update tested AGP versions" (inside it bumps 10.0.0-rc01 to rc02)
- "Update tested AGP versions" (inside it bumps 9.0.0 to 9.0.1)

I suggest removing this entire 'package rule' as its main purpose was to group them and the message meant to reflect that grouping.

.github/renovate.json5
```diff
@@ -53,13 +53,17 @@
...
+    {
+      // Ignore broken 9.2.0-rc01
+      "matchDepNames": ["com.android.tools.build:gradle"],
+      "allowedVersions": "!/^9\\.2\\.0-rc01$/",
     },
```
gabrielfeo: There's actually a simpler alternative to ignore a specific version that doesn't require adding such a specific rule. Once you remove the grouping and Renovate opens one PR for each distinct bump, you can close the PR for 9.2.0-rc01 which you want to ignore.

gabrielfeo (changes requested):

.github/renovate.json5
clayburn: Good point, thanks for raising it

gabrielfeo (approved):

